Issue #13, opened 2019-02-02
unexpected behaviors


matt-long:
I am starting to work with `intake_cesm`, but find that it does not behave as expected.

First point: the query function seems like it might not be working correctly. For instance, let's say I want to get data for the variable `FG_CO2` for experiments '20C' and 'RCP85'
```
>>> cat = collection.search(experiment=['20C', 'RCP85'], component='ocn', variable='FG_CO2')
```
Seems to work, but when I do 
```
>>> cat.results.files.tolist()
```
I see that the list includes files from the control integration (`experiment='CTRL'`):
```
['/glade/collections/cdg/data/cesmLE/CESM-CAM5-BGC-LE/ocn/proc/tseries/monthly/FG_CO2/b.e11.B1850C5CN.f09_g16.005.pop.h.FG_CO2.040001-049912.nc',
 '/glade/collections/cdg/data/cesmLE/CESM-CAM5-BGC-LE/ocn/proc/tseries/monthly/FG_CO2/b.e11.B1850C5CN.f09_g16.005.pop.h.FG_CO2.050001-059912.nc',
 '/glade/collections/cdg/data/cesmLE/CESM-CAM5-BGC-LE/ocn/proc/tseries/monthly/FG_CO2/b.e11.B1850C5CN.f09_g16.005.pop.h.FG_CO2.060001-069912.nc',
 '/glade/collections/cdg/data/cesmLE/CESM-CAM5-BGC-LE/ocn/proc/tseries/monthly/FG_CO2/b.e11.B1850C5CN.f09_g16.005.pop.h.FG_CO2.070001-079912.nc',
....
```
So that's unexpected.

Second issue, there is some subtlety in loading files into `xarray.Datasets`. 
```
ds = cat.to_xarray()
```
returns a dataset, but it appears to have concatenated *everything* along a time dimension! 
```
<xarray.Dataset>
Dimensions:               (d2: 2, lat_aux_grid: 395, moc_comp: 3, moc_z: 61, nlat: 384, nlon: 320, time: 96300, transport_comp: 5, transport_reg: 2, z_t: 60, z_t_150m: 15, z_w: 60, z_w_bot: 60, z_w_top: 60)
Coordinates:
  * lat_aux_grid          (lat_aux_grid) float32 -79.48815 -78.952896 ... 90.0
  * moc_z                 (moc_z) float32 0.0 1000.0 ... 525000.94 549999.06
[...]

PR #14, opened 2019-02-04
queries as list interpreted w OR; prototype of ensemble/variable loop…

matt-long:
This pull request partially address #13.

The changes to `_open_dataset` are an improvement, but this method remains fragile and incomplete.


intake_cesm/core.py
```diff
@@ -141,16 +141,44 @@ def results(self):
...
+                            ),
+                        )
+                    )
+
+                    ds_ens_list.append(dsi)
+
+            self._ds = xr.concat(ds_ens_list, dim='ens',
+                                 data_vars=variables)
+        else:
+            self._ds = _open_dataset(url, chunks=self.chunks, **kwargs)
```
andersy005: @matt-long, I think this needs to be changed to `self._ds = xr.open_dataset(url, chunks=self.chunks, **kwargs)`

matt-long: done.  Reopening PR

PR #17, opened 2019-02-04
build collection from multiple locations; bug fix for queries; reformat `_open_dataset`

matt-long:
This PR includes the following.
- renamed `collections.py` to `manage_collections.py` to avoid collision with `collections` python package.
- refactored `_build_cesm_collection` method that relies on list of `locations` and supports POSIX and HPSS file listings.
- bug fix in `common.py` that enable query args that are lists to be handled correctly.
- refactor of `_open_dataset` method in `core.py`: this remains in preliminary status.


matt-long:
Ok...this is a mess. I removed the csv files in collections because they were getting too big, but did not appreciate the dependence of testing on these files being present... I have rearranged so that the test operate only on the "cesm_dple" collection, which is small.  Also moved the input files out of the package directory.

matt-long:
@andersy005 can you take a look at this please?  I am out of time.

intake_cesm/manage_collections.py
```diff
@@ -0,0 +1,301 @@
...
+import logging
+import os
+import shutil
+import re
+
+from subprocess import Popen, PIPE
+
+import pandas as pd
+import yaml
+from tqdm import tqdm
```
andersy005: `tqdm` needs to be added to the list of requirements in `setup.py` by changing this line:

`install_requires = ["intake", "xarray", "pyyaml"]`

andersy005:
@matt-long, Thank you for pulling this off!

andersy005:
> Also moved the input files out of the package directory.

Because you moved the collection_dir out of the package directory, this directory would not be accessible when the user installs the package via `pip install intake-cesm`. As a result, we will need to make a couple of changes in `config.py`, `common.py`, `core.py` for the rest of the source code to work. 

1. We can get rid of `config.py` since we were doing very minimal work in that module. As you can see we are return a dictionary in `config.py`. 

2.  In`core.py`: we could just change the constructor of `CesmMetadataStoreCatalog` by adding an argument for an explicit path for the collection:

```python
class CesmMetadataStoreCatalog(Catalog):
    name = "cesm_metadatastore"
    version = __version__

    def __init__(self, collection, **kwargs):
         self.collection = collection
         self.df = open_collection(collection)
```

to 

```python
class CesmMetadataStoreCatalog(Catalog):
    name = "cesm_metadatastore"
    version = __version__

    def __init__(self, collection, collection_path, **kwargs):
          self.collection = collection
          self.collection_path = collection_path
          self.df = open_collection(collection_path)
```

3. `common.py`:

     - We need to remove this line `from .config import collections`

    - Change `open_collection()` to

```python
def open_collection(collection_path):
    """ Open a CESM collection and return a Pandas dataframe """
    try:

        df = pd.read_csv(collection_path, index_col=0)
        return df

    except (KeyError, FileNotFoundError) as err:
        print("****** The specified collection does not exit. ******")
        raise err
```

 - Change `get_subset()` to:

```python
def get_subset(collection_path, query):
    """ Get a subset of collection entries that match a query """
    df = open_collection(collection_path)
```

intake_cesm/manage_collections.py
```diff
@@ -0,0 +1,301 @@
...
+            else:
+                filelist.append(lines[i])
+                i +=1
+
+        return filelist
+
+
+
+class CESMCollections(object):
+    db_dir = "./collections"
```
andersy005: Having `db_dir` as a relative path might create problems during/after package installation. My recommendation would be to use a "stable" and absolute location path such as `~/.intake-cesm/collections`

andersy005: I would like take the first point in my previous comment back :)

> 1. We can get rid of `config.py` since we were doing very minimal work in that module. As you can see we are just returning a dictionary in `config.py`. 

We can keep `config.py` and use it for some package configurations such as creating the `~/.intake-cesm/collections` I mentioned above. 

```python
import os

HOME = os.environ.get("HOME", None)
db_dir = f"{HOME}/.intake-cesm/collections"
os.makedirs(db_dir, exists_ok=True)
```

With this change implemented, we could just import `db_dir` in `manage_collections.py`:
```python
from .config import db_dir
```

Let me know what you think.

matt-long: I have implemented a `SETTINGS` directory in `config.py`. This is settable via a `set_options` class, inspired by `xarray`. 

Test collection are now stored in tests/test_collections. The database_directory is set in tests/__init__.py as follows
```
set_options(database_directory='./test_collections')
```

This seems to be a good solution.

andersy005:
@matt-long, this looks good to me. Don't worry about the codecov patch check. I am going to go ahead and merge this.

matt-long:
Whew! Thanks!

PR #35, opened 2019-02-16
Code Refactoring : Allow CMIP integration

andersy005:
This PR does the following refactoring prior `intake-cmip` integration within `intake-esm`:

- [x] Move cesm_definitions to `config.py`
- [x] Add test_config.py
- [x] Rename `CesmMetadataStoreCatalog` to `ESMMetadataStoreCatalog` and add functionality for supporting `CMIP` integration.


andersy005:
@matt-long, when you get time, can you take a look this? I made quite a lot of changes and would like this to be merged once you give it a green light.

andersy005:
@matt-long, 

Because of `cmip` integration, I had to change the interface.

To build collections, you would need to run the following:
```python
col = intake.open_esm_metadatastore(collection_input_file=collection_type_def_file)
col.build_collections()
```
This returns an instance of `ESMMetadataStore` class from `intake_esm`. Once the collections are built, you would need to call `open_colllection()` method of this class by specifying the name of the `collection` as defined in `collection_input_file` and what type of collection this is: `cesm` or `cmip`. 

```python
col = col.open_collection(collection_name="cesm_dple", collection_type="cesm")
```

I am open to ideas on how to make this interface better.

matt-long:
@andersy005, this is an impressive amount of work and a great step forward. I like the solution of putting the catalog definitions in config.

I am concerned, however, about the complexity of this interface. Here a few observations.
- It seems most intuitive to me if there is a single call that enables opening, and optionally building the collection.
- The `collection_name` argument is problematic: at present, we have this in the input file and specified by the user—they might not match. Also, what if the user builds different collections in different projects with the same name?
- The CESM collection input file contains all the information defining the collection; we don't have a comparable input file for CMIP. 

I think it would be helpful to discuss in person.

andersy005:
> It seems most intuitive to me if there is a single call that enables opening, and optionally building the collection.

This sounds like an ideal approach. However, the issue arises when you have collections of different types in the `collection_input_file`. For instance, if you had a mixture of `cmip5` and `cesm` collections in the `collection_input_file`, what would you return after all the collections are built? When this input file contains homogeneous collections, interface you are proposing would definitely be the right solution. 


> The `collection_name` argument is problematic: at present, we have this in the input file and specified by the user—they might not match. Also, what if the user builds different collections in different projects with the same name?

I added a partial solution to this that makes it easier for the user. Once the collections are built, `self.collections` attributes of `ESMMetadataStore` is updated. This a dictionary of built collections and their corresponding paths. I use the same dictionary when loading built collections:

https://github.com/NCAR/intake-esm/blob/d36e84f2410fa4f696c346a5b9d4d2e6b1a5dc2d/intake_esm/core.py#L55-L60

https://github.com/NCAR/intake-esm/blob/d36e84f2410fa4f696c346a5b9d4d2e6b1a5dc2d/intake_esm/core.py#L87-L98

> The CESM collection input file contains all the information defining the collection; we don't have a comparable input file for CMIP.

We will need to put together a definition file for CMIP. 

> I think it would be helpful to discuss in person.

I am available this afternoon. Let me know, and I will stop by your office. If you have a lot on your place, I can stop by another time tomorrow

matt-long:
Regarding the issue with `type`:
- constrain the input file specification to allow one and only one collection.

Another idea: permit the "contents" of the input file to be passed in as arguments:
- name
- type
- data_sources
  - name
  - locations
  - component_attrs
  - case_members

The problem is we are "discovering" these things from the CMIP-AP holdings, while "specifying" them with the CESM input file. 

I am book solid today.

andersy005:
> constrain the input file specification to allow one and only one collection.

Sounds like a great constraint :). I will work on this 

> The problem is we are "discovering" these things from the CMIP-AP holdings, while "specifying" them with the CESM input file.

This won't be a problem since I separated the implementation for both collection types. For the time being, we can keep using the same approach we used in `intake-cmip` to discover files' attributes.

matt-long:
@andersy005, what's the status here?

andersy005:
@matt-long, this is still work in progress. I didn't get to work on it last week. I will address issues you pointed out some time tomorrow.

andersy005:
@matt-long, this is ready for another review whenever you have time

intake_esm/core.py
```diff
@@ -1,195 +1,151 @@
...
 
-    name = "cesm_metadatastore"
+
+class ESMMetadataStoreCatalog(Catalog):
+    """ESM collection Metadata store. This class servers as an entry point for `intake_esm`. """
+
+    name = 'esm_metadatastore'
     version = __version__
+    collection_types = {'cesm': CESMCollection, 'cmip': CMIPCollection}
+    SOURCES = {'cesm': 'intake_esm.cesm.CESMSource', 'cmip': 'intake_esm.esm.CMIPSource'}
```
matt-long: @andersy005: there is comparable `SOURCES` defined in config. Should it be in both places?

intake_esm/core.py
```diff
@@ -1,195 +1,151 @@
...
 
-    def build_collections(self):
-        """ Build CESM collection
         """
+        self.collections = {}
+        self.get_built_collections()
 
-        if "collection_input_file" in self.build_args:
-            collection_input_file = self.build_args.get("collection_input_file")
+        if collection_name and collection_type:
```
matt-long: What if `collection_name and collection_type` *and* `collection_input_file`?

intake_esm/core.py
```diff
@@ -1,195 +1,151 @@
...
+    def build_collection(self):
+        ctype = self.input_collection['collection_type']
+        cc = ESMMetadataStoreCatalog.collection_types[ctype]
+        cc = cc(self.input_collection)
+        cc.build()
+        self.get_built_collections()
+        self.open_collection(
+            self.input_collection['name'], self.input_collection['collection_type']
+        )
+        return self
```
matt-long: necessary?

intake_esm/core.py
```diff
@@ -1,195 +1,151 @@
...
 
-        Returns:
-            intake.Catalog -- An intake catalog entry
-        """
+    def open_collection(self, collection_name, collection_type):
+        """ Open an ESM collection"""
+        self.df, self.collection_name, self.collection_type = _open_collection(
+            collection_name, collection_type
+        )
+        return self
```
matt-long: again, is `return self` necessary?

intake_esm/core.py
```diff
@@ -1,195 +1,151 @@
...
 
-        kwargs.setdefault("name", collection)
-        super(CesmMetadataStoreCatalog, self).__init__(**kwargs)
-        if self.metadata is None:
-            self.metadata = {}
-        self._entries = {}
 
-    def build_collections(self):
-        """ Build CESM collection
         """
```
matt-long: Should the `kwargs` be documented? These appear to be passed thru to the `intake.catalog.Catalog` class. How do you envision this being used?

intake_esm/cesm.py
```diff
@@ -0,0 +1,342 @@
...
+                )
+
+        # Loop over ensemble members
+        for ensemble, ensemble_attrs in enumerate(ensembles):
+            input_attrs_base = {'experiment': experiment}
+
+            # Get attributes from ensemble_attrs
+            case = ensemble_attrs['case']
+
+            if 'ensemble' not in ensemble_attrs:
```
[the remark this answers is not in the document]
andersy005: Can you expand on this?

intake_esm/core.py
andersy005: Good catch! I will fix it

intake_esm/core.py
andersy005: Good point. How about raising an exception?

intake_esm/core.py
andersy005: I agree. This is not needed here

intake_esm/core.py
andersy005: Not needed here too. I will remove it

intake_esm/core.py
andersy005: I didn't have any plan on using them. So, I removed them and just added an optional argument `metadata` which is a dict of arbitrary information  to carry along with the data collection source specs.  If it's useless, we can remove it entirely :)

andersy005:
Thank you, @matt-long! I've addressed the issues you pointed out.

intake_esm/core.py
andersy005: I addressed this (probably not in the most elegant way :)). If you have suggestions for better alternative, let me know

intake_esm/cesm.py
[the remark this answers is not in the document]
matt-long: I was thinking: does it make sense to have a centralized data structure, something like `column_defaults` and just loop thru that?  (`ensemble` is a little different than the others.)

intake_esm/core.py
matt-long: I thought about this and haven't come up with a clear alternative. I could adding a class method for building the collection from a file, and pass thru to this method when we detect that a file name is provided...but it's hard to know that a string is actually a file. I don't want to require the user to pass in a file object.

PR #42, opened 2019-03-09
enable local config file

matt-long:


intake_esm/config.py
```diff
@@ -19,6 +19,10 @@
     PATH = os.environ['INTAKE_CONFIG']
     paths.append(PATH)
 
+elif os.path.exists(os.path.join(os.getcwd(), '.intake_esm')):
```
martindurant: I don't know if this makes sense to your use-case, but have you thought about config files which are not local? In most places in Intake, we attempt to allow for arbitrary URLs and "storage_options" for any parameters required to access remote storage.

matt-long: Thanks for the comment. I could see remote config files being useful in the context of defining new `collections`.  I don't have a current application, however.

Issue #43, opened 2019-03-10
CMIP6 version of cmip.py


naomi-henderson:
Thanks @andersy005 for your helpful additions to esgf2xarray !   I am also interested in combining `esgf_search` of `esgf2xarray` and `open_esm_metadatastore` of `intake-esm` similar to your description of discussion with @matt-long in  https://github.com/pangeo-data/esgf2xarray/issues/1#issuecomment-470763581.  In my case I would also like to be able to compare a esgf dataframe with a intake-esm dataframe to keep track of the new CMIP6 data as it appears on ESGF to trigger the download of any data of interest to our local machines.

So ...  I am hoping to configure the intake_esm cmip functionality to work with the new directory structure of the CMIP6 archive.  ESGF changed the directory structure and made other important changes - there is no longer an entity called "realm" in the path and "table_id" specifies both realm and frequency.  There is an important new "grid_label"  and we probably need to keep track of some additional information (checksums?).   

Before I dive in and make a new version of [cmip.py](https://github.com/NCAR/intake-esm/blob/master/intake_esm/cmip.py) and add a 'cmip6' collection to [config.yaml](https://github.com/NCAR/intake-esm/blob/master/intake_esm/config.yaml),  I thought I would check and see if anyone else is working on this?

matt-long:
@naomi-henderson I think you can go ahead and work on this. @andersy005, please chime in if you feel otherwise.

I think we can improve the generality in [cmip.py](https://github.com/NCAR/intake-esm/blob/master/intake_esm/cmip.py), so I would hope that CMIP5 and CMIP6 (and perhaps even [cesm.py](https://github.com/NCAR/intake-esm/blob/master/intake_esm/cesm.py)) can be handled by the same code. This might entail pushing things like `required_columns` into the configuration file. We could also do better separating the code generating a list of data resource from the code parsing it.

naomi-henderson:
I agree @matt-long .  Having looked more carefully, it doesn't seem worthwhile to write separate code for CMIP6 at least.  The CMIP6 directory structure (at least for now) is much more regular and it is possible to generate the dataframe by just walking down the directory and doing a little checking for required columns and duplicates before saving as csv - all easily done.

andersy005:
@naomi-henderson, I wanted to let you know that I am currently working on this in #46. Feel to free to chime in. I will post an update once #45 is merged.

andersy005:
Addressed in #45, and #46

PR #45, opened 2019-03-15
Improvements to existing functionality

andersy005:
- [x] Update `to_xarray()` method to allow users to pass keyword arguments to this method
- [x] Refactor `cmip.py` in preparation for cmip6 integration as proposed in #43


andersy005:
@matt-long, I've experienced difficulties while trying to merge/concat CMIP data from different institutions. It appears that there are discrepancies among **coordinates** of different models' outputs. For queries in which there's data from more than one institution, xarray is unable to align the coordinates. 

For instance, I saw one variable in which the coordinates in one model are `time, lat, lon, bnds` and `time, rlat, rlon, bnds` in another model from a different institution. In addition to 

https://github.com/NCAR/intake-esm/blob/67fb828a202a7c7fc15953b63436075554e3ae1c/intake_esm/cmip.py#L240-L244

should we add a check that forbids merging datasets from different institutions? Let me know what would be the right approach.

matt-long:
@andersy005, I would not expect models from different institutions to be concatenate-able; the different models do indeed have different coordinates. I think we should add a check that ensures prevent this. Another approach would be to return a data structure that includes the datasets from the different models. A dictionary with the institution ID as the key, for instance, could work. This would entail an outer loop, probably best implemented as a method that calls this `_open_dataset` method.

andersy005:
@matt-long, this is ready for another look

intake_esm/aggregate.py
```diff
@@ -188,13 +197,13 @@ def set_coords(ds, varname):
     return ds.set_coords(coord_vars)
 
 
-def open_dataset(url, data_vars, chunk_size=default_chunk_size):
+def _open_dataset(url, data_vars, chunk_size=config.get('default_chunk_size'), **kwargs):
```
matt-long: I think this may lead to misleading behavior. A user might expect modifying config to modify the defaults, but since the API is defined at import-time, this won't happen. I suggest `chunk_size=None` here and 
```python
if chunk_size is None:
  chunk_size = config.get('default_chunk_size')
```
in the function.

intake_esm/aggregate.py
```diff
@@ -103,20 +95,23 @@ def concat_time_levels(dsets):
...
+    time_coord_name_default : string
+        Default name of the time coordinate
+
     Returns
     -------
     dset : xarray.Dataset,
         The concatenated dataset.
     """
-
+    dsets = dask.compute(*dsets)
```
matt-long: Does this belong here or up a level in the module that is calling `open_dataset`? It seems like it would at least be prudent to check if `dsets` actually contains `dask.delayed` objects.

intake_esm/aggregate.py
```diff
@@ -188,13 +197,13 @@ def set_coords(ds, varname):
...
     with dask.config.set({'array.chunk-size': chunk_size}):
-        ds = xr.open_dataset(url, chunks={'time': 'auto'}, decode_times=False)
+        ds = xr.open_dataset(url, chunks={'time': 'auto'}, **kwargs)
     ds.attrs['history'] = f"{datetime.now()} xarray.open_dataset('{url}')"
 
     return set_coords(ds, data_vars)
 
 
-open_dataset_delayed = dask.delayed(open_dataset)
+open_dataset = dask.delayed(_open_dataset)
```
matt-long: Do we want to preserved a public method for non-delayed access? I don't really understand the performance/robustness issues here. I am happy assuming `dask` in general, but should we insist on `dask.delayed`?

intake_esm/cesm.py
```diff
@@ -41,7 +41,7 @@ def __init__(self, collection_spec):
         self.df = pd.DataFrame(columns=self.columns)
 
     def _validate(self):
-        for req_col in ['files', 'sequence_order']:
+        for req_col in ['file_fullpath', 'sequence_order']:
```
matt-long: Should there be a `required_columns` in the catalog definition?

intake_esm/cesm.py
```diff
@@ -303,33 +303,18 @@ class CESMSource(BaseSource):
...
-        ).files.tolist()
+    def __init__(self, collection_name, collection_type, query={}, **kwargs):
+
+        super(CESMSource, self).__init__(collection_name, collection_type, query, **kwargs)
         self.query_results = get_subset(
             self.collection_name,
             self.collection_type,
             self.query,
-            order_by=['sequence_order', 'files'],
+            order_by=['sequence_order', 'file_fullpath'],
```
matt-long: Should there be an `order_by_columns` in the definition?

intake_esm/cesm.py
```diff
@@ -342,51 +327,70 @@ def results(self):
...
-                ds_ens_list.append(ds_ens_i)
-
-            self._ds = aggregate.concat_ensembles(ds_ens_list, member_ids=ensembles, join='outer')
-        else:
-            self._ds = xr.open_dataset(url, chunks=self.chunks, **kwargs)
+
+        if 'decode_times' not in kwargs.keys():
+            kwargs.update(decode_times=False)
+        if 'time_coord_name' not in kwargs.keys():
+            kwargs.update(time_coord_name='time')
```
matt-long: Should this be a call to `infer_time_coord_name`?

intake_esm/cesm.py
```diff
@@ -342,51 +327,70 @@ def results(self):
...
+                    self.collection_name,
+                    self.collection_type,
+                    query,
+                    order_by=['sequence_order', 'file_fullpath'],
+                )['file_fullpath'].tolist()
+
+                dsets = [
+                    aggregate.open_dataset(
+                        url, data_vars=[var_i], decode_times=kwargs['decode_times']
+                    )
```
matt-long: What about the chunk argument?

intake_esm/aggregate.py
```diff
@@ -179,6 +181,13 @@ def concat_ensembles(dsets, member_ids=None, join='inner'):
         attrs['history'] = new_history
     ds.attrs = attrs
 
+    # rechunk
+    if chunks is None:
```
matt-long: We have discussed potential performance implications of re-chunking. Since we are concatenating along a new dimension, it seems like in theory we could avoid re-chunking along time and impose chunks during the concat (though I am not sure that supported by xarray).

intake_esm/cmip.py
```diff
@@ -225,31 +213,71 @@ def results(self):
...
+        query = dict(self.query)
+
+        if 'decode_times' not in kwargs.keys():
+            kwargs.update(decode_times=False)
+        if 'time_coord_name' not in kwargs.keys():
+            kwargs.update(time_coord_name='time')
+        if 'ensemble_dim_name' not in kwargs.keys():
+            kwargs.update(ensemble_dim_name='member_id')
+        if 'chunks' not in kwargs.keys():
+            kwargs.update(chunks=None)
```
matt-long: This code is duplicate that in `cesm.py`. Are there ways we can avoid this duplication? As things evolve, identical implementations might be required in two places.

intake_esm/cmip.py
```diff
@@ -225,31 +213,71 @@ def results(self):
...
-        self._ds['ensemble'] = ens_list
+        if len(frequency_list) > 1:
+            raise ValueError(
+                f'Found multiple data frequencies: {frequency_list} in query results. Please specify the frequency to use'
+            )
+
+        _ds_dict = {}
+        grouped = get_subset(self.collection_name, self.collection_type, query).groupby(
+            'institution'
+        )
```
matt-long: Does the code above belong in a `_validate_concat_open_dataset` method?

intake_esm/cmip.py
```diff
@@ -225,31 +213,71 @@ def results(self):
...
+        for name, group in grouped:
```
matt-long: Should we have an `_open_dataset_groups` method that calls the `_open_dataset` method? Would this allow us to reuse more code between `cesm.py` and `cmip.py`?

intake_esm/config.yaml
```diff
@@ -117,9 +117,9 @@ collections:
       - ocean
       - ocnbgchem
       - seaIce
+default_chunk_size: 12MiB
```
matt-long: I don't know if this is a good value. Should we stick with the `dask` default as our default?

intake_esm/cesm.py
andersy005: At this stage, we don't have access to any xarray dataset yet. So, `infer_time_coord_name()` wouldn't work. Even though the default value here is `time`, in the next steps, I call `infer_time_coord_name()` to confirm that this is the right time coordinate name prior doing any operation

intake_esm/cesm.py
andersy005: This was a tricky case to handle due to the fact that a user may provide `chunks` argument with dimensions that are not available at this stage. For instance, if the user provides `chunks={'member_id' : 2}`, and we try to pass this to `aggregate.open_dataset()`, this function will raise an error due to absence of `member_id` in the dataset. Because of this, I decided to wait until after the final concatenation to chunk the dataset into desired chunks when we are 100% sure that all dimensions are present.

intake_esm/config.yaml
andersy005: Changed it to dask default of `128MiB`

intake_esm/aggregate.py
andersy005: > Do we want to preserved a public method for non-delayed access?

I don't think this is giving us that much performance gain, if any.  I've disabled it for the time being

intake_esm/cmip.py
andersy005: I see the advantage of generalizing this. What would be the equivalent of `_open_dataset_groups()` for cesm data, in other words, what is the equivalent of CMIP institutions in CESM?

matt-long: We could think about this as moving toward any arbitrary collection of datasets that should not be concatenated along a `dimension`.  Observations, integrations from different model version or resolutions, for instance. We might want to have a model general `dataset_id` or something versus insisting on "institution."

intake_esm/aggregate.py
andersy005: I just realized that we may have a slight problem. One line 204, we are using `time` as the default time coordinate, and I am wondering whether there is any way to replace this assumption with something robust?

https://github.com/NCAR/intake-esm/blob/ef62f70610519f1197e5ed975d2f290d1440375d/intake_esm/aggregate.py#L201-L207

intake_esm/cesm.py
matt-long: Fair enough. Should `infer_time_coord_name` be renamed to `ensure_time_coord`?

intake_esm/cmip.py
andersy005: I will take a stab at generalizing it (I am afraid that it may not a trivial task though)

intake_esm/aggregate.py
matt-long: We're expecting `chunks` in the args `_open_dataset`. Do we need to not chunk here because we're chunking later in `concat_ensembles` or do we want to subset `kwargs['chunks']` for the
`time_coord_name` entry?

intake_esm/cesm.py
```diff
@@ -342,51 +320,57 @@ def results(self):
...
+                    for url in urlpath_ei_vi
+                ]
+                ds_var_i = aggregate.concat_time_levels(dsets, kwargs['time_coord_name'])
+                ds_var_list.append(ds_var_i)
+
+            ds_ens_i = aggregate.merge(dsets=ds_var_list)
+            ds_ens_list.append(ds_ens_i)
+
+        self._ds = aggregate.concat_ensembles(
+            ds_ens_list, member_ids=ensembles, join='outer', chunks=kwargs['chunks']
```
matt-long: I think the user needs control over `join`; we should enable this being passed in.

intake_esm/aggregate.py
andersy005: Not 100% sure, I have the right approach in the new changes. Please check the new changes out and let me know what you think.

matt-long: this seems reasonable.

intake_esm/cmip.py
andersy005: Could you expand on this?

matt-long: I think it's probably fine for now. I was wondering whether you could write a general method for validation. Not sure of the API....
```python
def _validate_concat(check_max_instance=None):

  if check_max_instance is not None:
    for fld, max_inst in check_max_instance.items():
        fld_list = self.query_results[fld].unique()
        if len(fld_list) > max_inst:
            raise ValueError(f'message about {fld}')
```

intake_esm/cesm.py
andersy005: Done!

intake_esm/cmip.py
andersy005: I see! Now that we are handling an addition case of  datasets that cannot be concatenated by returning a dictionary of dsets in:

https://github.com/NCAR/intake-esm/blob/eb8fdafa0d7b22770bffde968bf36453d9cd5b9a/intake_esm/cmip.py#L254-L264

can you think of other cases that wouldn't be handled by the above code?

matt-long: I probably need to work thru some use cases to really develop my intuition here.

andersy005: Sounds good! Let me know when you get to play with this.

Issue #66, opened 2019-05-06
Extending intake-esm to RDA holdings: add support for ERA5 Reanalysis


andersy005:
https://rda.ucar.edu/datasets/ds630.0/

Resources/Documentation:
 - https://confluence.ecmwf.int/pages/viewpage.action?pageId=85402030#ERA5terminology:analysisandforecast;timeandsteps;instantaneousandaccumulatedandmeanratesandmin/maxparameters-Instantaneous,accumulated,meanrateandmin/maxparameters
- https://confluence.ecmwf.int/display/CKB/ERA5+data+documentation#ERA5datadocumentation-Dataorganisationandaccess
- https://apps.ecmwf.int/codes/grib/param-db

andersy005:
Addressed by #68

mickaellalande:
Is ERA5 reanalysis accessible through intake? Is there an example in the documentation of how to access it?

wachsylon:
@andersy005 

I made one for dkrz era5 data which is on our HPC's disk storage. That means, it also only works on Mistral.

```
dkrz_cdp=intake.open_catalog("https://swift.dkrz.de/v1/dkrz_a44962e3ba914c309a7421573a6949a6/intake-esm/dkrz_data-pool_cloudcatalog.yaml")
esm_dkrz_era=dkrz_cdp.dkrz_era5_disk_grb_fromcloud
```
should work.

However, it is based on raw `grb` data and `intake-esm` does not really support `grb`. We would need a default engine [here](https://github.com/intake/intake-esm/blob/40fe3a7808c14af77d96cfc932b9cebaa6beba10/intake_esm/source.py#L18) for grib. The one i use is `cfgrib`. You may say that users can provide that but since there is no `grib_kwargs`, I cannot for example merge `grb` assets with `netcdf` assets. Which is why I say there is no `grb` support in intake-esm.

Best,
Fabi

andersy005:
@wachsylon, I concur. We should expand the specification to allow intake-esm catalogs to contain `grib`: https://github.com/intake/intake-esm/blob/main/docs/source/explanation/esm-collection-spec.md#assets-object

andersy005:
@wachsylon, I opened an issue for this: https://github.com/intake/intake-esm/issues/452

PR #68, opened 2019-05-08
Add support for ERA5 Reanalysis

andersy005:
Addresses #66


martindurant:
I wonder, do you have a master catalogue pointing at all of these resources that you are now able to load? I am thinking along the lines of https://github.com/pangeo-data/pangeo-datastore/blob/master/intake-catalogs/master.yaml - and note that that repo also has CI to ensure that all entries load OK, and renders the whole cat to a static webpage on mergers too.

(PS: I think you must have included notebook outputs, for the diff to be so large)

andersy005:
@martindurant, we don't have a master catalogue yet. We will eventually have a master catalogue for community datasets such as CMIP and CESM-LE. Thank you for the pointer to the pangeo datastore repo ( I wasn't aware of it). The continuous integration + Browsable website is really great. I will see if there's a possibility of adapting a  similar approach to our catalogues. 


> (PS: I think you must have included notebook outputs, for the diff to be so large)

I found the culprit to be this file:
https://github.com/andersy005/intake-esm/blob/3684f9329598f9eb0e841d0096fbfed27cfa3dd0/tests/era5_filelist.txt

PR #85, opened 2019-06-17
ensemble-->member_id

matt-long:


andersy005:
This line https://github.com/NCAR/intake-esm/blob/master/intake_esm/cesm.py#L92 needs to be updated as well.

PR #120, opened 2019-08-07
Made experiment independent dataset field for cordex data

jukent:
Intake-esm automatically concatenates time for historical and rcp85 or other scenarios which is not ideal. Encoding information is lost in this concatenation.

### DIFF
--- a/intake_esm/cordex.py
+++ b/intake_esm/cordex.py
@@ -49,7 +49,7 @@ class CORDEXSource(BaseSource):
             'frequency',
             'grid',
             'bias_corrected_or_raw',
-            'experiment'
+            'experiment',
         ]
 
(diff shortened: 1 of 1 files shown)


andersy005:
@jukent, 

When you get time, can you fix the linting issue (by installing and running the pre-commit hooks) and then we can go ahead and merge this (unless you are still working on it)?

jukent:
@andersy005 For some reason pre-commit is already installed, but when I run pre-commit install && pre-commit run it says pre-commit command cannot be found.

andersy005:
> but when I run pre-commit install && pre-commit run it says pre-commit command cannot be found.

it's likely that you are running the command from a different conda environment other than the one in which `pre-commit` is installed in..

Check the output of `conda list` prior running the `pre-commit` commands

PR #154, opened 2019-10-17
Run tests against xarray's master branch

### LINKED ISSUE #149: test against xarray master
It would be mutually beneficial to test intake-esm against xarray master so that we know when we are unintentionally breaking things and you know when things are about to be deprecated before an actual release.

andersy005:
Closes #149


dcherian:
woot! Thanks @andersy005

Issue #152, opened 2019-10-16
User-defined merge operation


rabernat:
One issue that has emerged from the hackathon is the desired to apply user-defined preprocessing and / or customized merging function during the dataset loading process.

@jbusecke has a great example of this. He wants to use xgcm to calculate derivatives, divergence, etc. This requires that the data variables on a staggered grid model (e.g. C-grid, B-grid) be labeled with different dimensions depending on their grid position (e.g cell center, `lon_c`, cell face, `lon_g`, etc.) We have found that many (all) of the CMIP6 data just use the same dimension names (e.g. lon, lat) for all variables (e.g. theatao, uo, vo), regardless of their grid position. However, this information is implicitly encoded in the actual coordinate values, which are different. (Related to #151.)

Currently, if you use intake-esm to merge, say, `thetao` and `uo`, you will end up doubling the length of each variable and inserting missing values, due to the staggered nature of the coordinates. Instead, what we want to do is relabel these as distinct dimensions.

Julius has some code to do this here:
https://github.com/jbusecke/cmip6_preprocessing/blob/master/cmip6_preprocessing/recreate_grids.py

This is similar to function I wrote a while back for POP:
https://github.com/jbusecke/cmip6_preprocessing/blob/master/cmip6_preprocessing/recreate_grids.py

**How would we plug this sort of custom merge operation into intake-esm?**

Would it be possible to provide a callback for a custom merge function? If so, what API would such a function have to implement? Can we define an interface for this?

cc @naomi-hendersion

matt-long:
I think it may be relatively easy to implement a `preprocess` argument to `to_dataset_dict` that would accept a function that operates on the individual datasets as they are opened (could also accept kwargs). It could do something like this

```python
def fix_coords(ds):
  if "vo" in ds.data_vars or "uo" in ds.data_vars:
    # rename coords/dims lon-->lon_u etc.
 ```
and the user would do something like this
```python
dsets_dict = col.to_dataset_dict(preprocess=fix_coords)
```
This doesn't get all the way to an `xgcm` compatible dataset, but would prevent `xarray` from interleaving the coordinates and the full `xgcm`-ification that @jbusecke is doing could be applied to the returned, merged dataset (unless I've missed something). This should be easy to implement.

I think the custom merge is more complicated. We will need to think through and generalize how the merge-function gets information from the collection, i.e. which dataset is which, etc.

jbusecke:
Testing now.

jbusecke:
[...]

PR #155, opened 2019-10-17
add preprocess 

matt-long:
This PR enables a `preprocess` argument to `to_dataset_dict`. For example:
```python
    def rename_coords(ds):
        return ds.rename({'lon': 'longitude', 'lat': 'latitude'})
    
    col = intake.open_esm_datastore(esmcol_path)
    col_sub = col.search(**query)
    
    dsets = col_sub.to_dataset_dict(preprocess=rename_coords)
```
cc @rabernat @jbusecke

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -237,9 +237,9 @@ class esm_datastore(intake.catalog.Catalog, intake_xarray.base.DataSourceMixin):
             Keyword arguments to pass to `xarray.open_zarr()` function
         cdf_kwargs : dict
             Keyword arguments to pass to `xarray.open_dataset()` function
-        preprocess : (callable, optional) 
+        preprocess : (callable, optional)
             If provided, call this function on each dataset prior to aggregation.
-            
+
         Returns
         -------
         dsets : dict
@@ -322,12 +322,12 @@ class esm_datastore(intake.catalog.Catalog, intake_xarray.base.DataSourceMixin):
 
         self.zarr_kwargs = zarr_kwargs
         self.cdf_kwargs = cdf_kwargs
-        
+
         if preprocess is not None and not callable(preprocess):
             raise ValueError('preprocess argument must be callable')
-            
+
         self.preprocess = preprocess
-        
+
         return self.to_dask()
 
     def _get_schema(self):
@@ -457,7 +457,15 @@ def _load_group_dataset(
         )
 
     ds = aggregate(
-        aggregation_dict, agg_columns, n_agg, nd, lookup, mapper_dict, zarr_kwargs, cdf_kwargs, preprocess,
+        aggregation_dict,
+        agg_columns,
+        n_agg,
+        nd,
+        lookup,
+        mapper_dict,
+        zarr_kwargs,
+        cdf_kwargs,
+        preprocess,
     )
     group_id = '.'.join(key)
     return group_id, _restore_non_dim_coords(ds)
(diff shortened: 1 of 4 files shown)


andersy005:
Towards #152

jbusecke:
This functionality does not work as expected for me: 

Quick example:
```python
import intake
query = dict(experiment_id='historical', table_id='Oyr', 
                 variable_id='o2', grid_label='gn', member_id='r1i1p1f1', source_id='CESM2-WACCM')
cat = col.search(**query)

def test_func(ds):
    if 'nlon' in ds.dims:
        ds = ds.rename({'nlon':'WAZOOOOOONGA'})
    return ds

dsets_pp = cat.to_dataset_dict(preprocess=test_func)

dsets_pp['CMIP.NCAR.CESM2-WACCM.historical.Oyr.gn'].dims
```
gives `Frozen(SortedKeysDict({'nlat': 384, 'nlon': 320, 'vertices': 4, 'time': 165, 'd2': 2, 'lev': 60, 'member_id': 1}))
`
When I apply the function straight to the dataset I get this:
```
test_func(dsets_pp['CMIP.NCAR.CESM2-WACCM.historical.Oyr.gn']).dims
```

I get the expected change `Frozen(SortedKeysDict({'d2': 2, 'lev': 60, 'member_id': 1, 'nlat': 384, 'WAZOOOOOONGA': 320, 'time': 165, 'vertices': 4}))
`

Am I missing something or is this a bug?

matt-long:
@jbusecke, we are not able to reproduce this error. Are you using the most recent collection spec from 
https://github.com/NCAR/intake-esm-datastore/tree/master/catalogs
?

jbusecke:
No. Let me try again in a minute.

jbusecke:
I updated to the latest and it still produces the same error. See here.
https://github.com/jbusecke/cmip6_preprocessing/blob/master/notebooks/testing_intake_esm_preprocessin.ipynb

andersy005:
> Is the version not the latest?

Yep

jbusecke:
I updated the notebook to double check that there is no error on my side. Could you pull it and try to run it on your side?

https://nbviewer.jupyter.org/github/jbusecke/cmip6_preprocessing/blob/master/notebooks/testing_intake_esm_preprocessin.ipynb

matt-long:
Ok, I think we have this pinned down.

You are calling  `to_dataset_dict` twice, first without the `preprocess` argument. On the second call when you pass this in, however, `intake-xarray` (under the hood) is simply returning the cached dataset on the object: it is not detecting the change in the input args to appropriately update output. 

I am fixing this behavior now.

rabernat:
It would be ideal if `to_dataset_dict` should not have [side effects](https://en.wikipedia.org/wiki/Side_effect_(computer_science)).

jbusecke:
Thanks for the quick fix @andersy005, I am now able to use custom preprocessing functions! 

I made a quick example [here](https://github.com/jbusecke/cmip6_preprocessing/blob/master/notebooks/testing_intake_esm_preprocessin.ipynb), should we include that in the docs/readme?

naomi-henderson:
@jbusecke and @andersy005 , this is a great example for the documentation! 

By the way,  which `col' are you using, cloud or glade?  It doesn't seem to be specified - but you do call `to_dataset_dict` without adding the `zarr_kwargs` or `cdf_kwargs`.  If it is `pangeo-cmip6.json`, not using `zarr_kwargs={'consolidated': True}` is very slow.

Since your example uses your `cmip6_preprocessing` package, it would be good to point to your repo or conda/pip install method.  How about using a simple pre-processing function defined in the notebook itself, instead?

jbusecke:
See #168

PR #158, opened 2019-10-17
do not cache datasets

matt-long:


andersy005:
Addresses https://github.com/NCAR/intake-esm/pull/155#issuecomment-543327776

PR #180, opened 2019-11-14
Allow passing extra storage options to the backend file system

bonnland:


andersy005:
Thank you, @bonnland!

PR #194, opened 2020-01-14
Move logic for assets aggregation into `ESMGroupDataSource`

### LINKED ISSUE #165: Extract catalog entries by key
In vanilla Intake and Intake-xarray, users access datasets through the Entry class and the `to_dask` (or `to_xarray`), e.g:

```python
import intake
cat = intake.open_catalog('catalog.yml')
cat['key1'].to_dask()
```

Or for nested catalogs, 

```python
cat['key1.key2'].to_dask()
```

It would be nice if single aggregated datasets can be accessed in the same pattern:

```python
cat['CMIP.CCCma.CanESM5.historical.Ofx.gn'].to_xarray()
# is equivalent to
cat.to_datasets_dict(...)['CMIP.CCCma.CanESM5.historical.Ofx.gn']
```

It may be worth thinking about using the Entry class more explicitly to allow for more intake-like behavior in this way.

andersy005:
Summary: this PR will fix #165 and #226  

I have a binder set up in case you want to try it out [Binder Link](https://mybinder.org/v2/gist/andersy005/55917162f65ac88aebc9df83e519158c/master)

## TO DO

- [x] Create `ESMGroupDataSource` class and move assets aggregation logic into it
- [x] Populate (lazily) `catalog._entries` dictionary during the catalog class instantiation
- [x] Refactor `intake_esm.core.esm_datastore.to_dataset_dict`: Loop through items and invoke `source.to_dask()`

Tasks that will be addressed in separate PRs:
- Handling `aggregate=False`
- Updating the documentation
- Implementing `__repr_html__()` / `ipython_display`
- Remove some `intake-esm` keys added to datasets during aggregations from the global dataset attributes.


andersy005:
So far, here's what I have got:

```python
In [1]: import intake                                                                                                                        

In [2]: col_url = "https://raw.githubusercontent.com/NCAR/intake-esm-datastore/master/catalogs/pangeo-cmip6.json"                            

In [3]: col = intake.open_esm_datastore(col_url)                                                                                             

In [4]: len(col)                                                                                                                             
Out[4]: 227273

In [5]: # __getitem__ in action                                                                                                              

In [6]: # Full Key                                                                                                                           

In [7]: x = col['CMIP.CCCma.CanESM5.historical.r1i1p1f1.Ofx.deptho.gn']                                                                      

In [8]: len(x)                                                                                                                               
Out[8]: 1

In [9]: # Partial Key                                                                                                                        

In [10]: y = col['CMIP.CCCma.CanESM5.historical']                                                                                            

In [11]: len(y)                                                                                                                              
Out[11]: 4531

In [12]: type(x), type(y)                                                                                                                    
Out[12]: (intake_esm.core.esm_datastore, intake_esm.core.esm_datastore)
```

jhamman:
@andersy005 - this is looking pretty cool and I think you are getting close to what I was originally suggesting. 

> - col['CMIP.CCCma.CanESM5.historical.r1i1p1f1.Ofx.deptho.gn'] should return a new object Entry and

Right. You wil probably need to subclass `LocalCatalogEntry` and use logic in intake-esm to construct the xarray object when the `to_xarray()` object is called.

> - col['CMIP.CCCma.CanESM5.historical'] should return intake_esm.core.esm_datastore, right?

Yep. One question to consider is: what to do with the first four keys in the catalog. After indexing into the catalog in this way, you know everything will match `'CMIP.CCCma.CanESM5.historical'` so you might say you don't need this anymore. If this were an actual intake catalog, you would basically have done: `col['CMIP']['CCCma']['CanESM5']['historical']` so there wouldn't be any need for the parent catalogs.

> - The new Entry object will behave differently from the original intake_esm.core.esm_datastore object i.e. once I have x = col['CMIP.CCCma.CanESM5.historical.r1i1p1f1.Ofx.deptho.gn'], I should be able to call ds = x.to_xarray() and get an xarray dataset.

Exactly. Ideally, `to_dataset_dict()` on the `esm_datastore` becomes a collection of calls to `to_xarray` in this Entry object.

tests/test_core.py
```diff
@@ -209,6 +209,22 @@ def test_to_dataset_dict_chunking_2(chunks, expected_chunks):
...
+    'CMIP.CCCma.CanESM5.historical.r1i1p1f1.Ofx.deptho.gn',
+    'DCPP',
+]
+
+
+@pytest.mark.parametrize('key', keys)
+def test_get_item(key):
+    col = intake.open_esm_datastore(zarr_col_pangeo_cmip6)
+    assert isinstance(col.df, pd.DataFrame)
+    assert len(col) >= 1
```
kmpaul: This doesn't actually seem to test `__getitem__`.  Where is `key` used?

kmpaul:
@jhamman If this:

> ... `col['CMIP']['CCCma']['CanESM5']['historical']` ...

is what Intake does normally, shouldn't that be the behavior that we emulate?  That is, instead of a single key that is split on `'.'`?  My guess is that your answer probably has to do with the use case:

```python
col['CMIP']['*']['*']['historical']
```

which wouldn't work as written because:

```python
col['CMIP']['*']['*'] == col['CMIP'] == True
```

Correct?

intake_esm/core.py
```diff
@@ -120,6 +120,29 @@ def search(self, **query):
...
+            key_parts = key.split('.')
+        else:
+            key_parts = [key]
+        N = len(columns)
+        # Pad key list for missing values so that len(key) == len(columns)
+        key_parts += ['*'] * (N - len(key_parts))
+        query = dict(zip(columns, key_parts))
+        # Filter out the query for valid entries
+        filtered_query = {k: v for k, v in query.items() if v != '*'}
+        return self.search(**filtered_query)
```
kmpaul: @andersy005 You might also check to see if the `len(self.search(**filtered_query)) == 1` and then return an `Entry` right here (i.e., in `__getitem__`).  I think that is appropriate to do that in `__getitem__` and leave the return value of `self.search(...)` unchanged.

jhamman:
> @jhamman If this:
> 
> > ... `col['CMIP']['CCCma']['CanESM5']['historical']` ...
> 
> is what Intake does normally, shouldn't that be the behavior that we emulate? That is, instead of a single key that is split on `'.'`? 

Intake supports both of these formats so I think we should aspire to do the same, so far as it is practical. 

> ```python
> col['CMIP']['*']['*']['historical']
> ```

Right, this wouldn't work and is a good place to draw the line between item selection and search. Intake catalogs support nesting. One view of an intake-esm catalog could be as a highly nested catalog but our typical search patterns drop this nesting.

intake_esm/core.py
```diff
@@ -120,6 +120,29 @@ def search(self, **query):
...
+
+    def __len__(self):
+        return len(self.df)
+
+    def __keys__(self):
+        ...
```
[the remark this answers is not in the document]
andersy005: Decided to remove it for the time being.

tests/test_core.py
andersy005: I updated the tests.

intake_esm/core.py
```diff
@@ -120,6 +120,29 @@ def search(self, **query):
         ret.df = self._get_subset(**query)
         return ret
 
+    def __getitem__(self, key):
+        path_column_name = self._col_data['assets']['column_name']
+        columns = self.df.columns.tolist()
+        columns.remove(path_column_name)
+
+        if isinstance(key, str) and '.' in key:
+            key_parts = key.split('.')
```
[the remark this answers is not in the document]
andersy005: Great idea 👍. This is now supported.

jhamman:
I'm trying this out now, will report back with a detailed review soon.

intake_esm/core.py
```diff
@@ -82,6 +82,66 @@ def __init__(self, esmcol_path, progressbar=True, log_level='CRITICAL', **kwargs
         self.cdf_kwargs = None
         self.preprocess = None
         self.aggregate = None
+        self.sep = sep
+
+    def __len__(self):
+        return len(self.df)
```
jhamman: Thinking about this a bit, I think this should be the number of datasets, not the number of rows in the catalog_file. Today, this would be essentially `len(self.to_dataset_dict())` -- but I have more comments on this point.

intake_esm/core.py
```diff
@@ -82,6 +82,66 @@ def __init__(self, esmcol_path, progressbar=True, log_level='CRITICAL', **kwargs
...
+
```
jhamman: We should add a `__keys__()` method to reflect which datasets are available.

intake_esm/core.py
```diff
@@ -82,6 +82,66 @@ def __init__(self, esmcol_path, progressbar=True, log_level='CRITICAL', **kwargs
...
+            data_format = asset[1][1]
+
+            if data_format == 'zarr':
+                driver = 'zarr'
+
+            else:
+                driver = 'netcdf'
+
+            # TODO: figure out how to pass other args such as storage_options, chunks, etc
+            entry = LocalCatalogEntry(
```
jhamman: I've backtracked a bit on my original design target here. What I'm thinking would be most useful (as a start) is to provide access to an intake Entry for each key typically found in `to_dataset_dict()`. So these two datasets would be equivalent:

```python
dset_dict = cat.to_dataset_dict()
ds = dset_dict['CMIP.CCCma.CanESM5.historical.Oyr.gn']

# and
cat['CMIP.CCCma.CanESM5.historical.Oyr.gn'].to_dask()
```

Then `to_dataset_dict()` becomes just a convenience method:

```python
def to_dataset_dict(self, **kwargs):
    return {key: self[key](**kwargs).to_dask() for key in self.keys()}
```

What we need right now is to understand how to take a group of rows from the catalog's dataframe, and construct a single `GroupedEntry` where the loading logic is implemented in intake-esm.  @andersy005, can you spec out what information this `GroupedEntry` would need to construct a single xarray Dataset?

andersy005:
@jhamman, 

When you get a chance, this PR is ready for another round of review

intake_esm/core.py
```diff
@@ -58,31 +57,201 @@ class esm_datastore(intake.catalog.Catalog):
...
+            'format_column_name': format_column_name,
+        }
+        return info
+
+    @property
+    def keys(self):
+        keys = list(map(lambda x: self.sep.join(x), self._keys))
+        return keys
+
+    def _load(self):
```
andersy005, line 170: Constructing intake catalog entries appears to be very expensive:

Screen Shot 2020-04-07 at 2 55 18 PM

This is a huge bottleneck for large intake-esm catalog. 

@jhamman, let me know if you have any ideas on how to proceed.

jhamman: @andersy005 - can you show a similar profile for a single call to `LocalCatalogEntry`?

andersy005: >  can you show a similar profile for a single call to LocalCatalogEntry?

Here's one:

Screen Shot 2020-04-07 at 6 36 46 PM
Screen Shot 2020-04-07 at 6 36 58 PM

jhamman: @andersy005 - thanks but I think we're missing some of the screen shot.

andersy005: I omitted the doc string part:

Screen Shot 2020-04-07 at 10 53 41 PM

jhamman: Thinking about this a bit, I guess I'm not particularly surprised. We're creating 7k+ `LocalCatalogEntries`. Is the data source pulling some metadata at this stage? Would it be reasonable to make the construction of these entries lazy?

andersy005: Thank you for the suggestion! Making the construction lazy improves the performance:

```python
In [1]: import intake                                                       

In [2]: url = "https://raw.githubusercontent.com/NCAR/intake-esm-datastore/m
   ...: aster/catalogs/pangeo-cmip6.json"                                   

In [3]: %%time 
   ...: col = intake.open_esm_datastore(url) 
   ...:  
   ...:                                                                     
CPU times: user 1.87 s, sys: 283 ms, total: 2.15 s
Wall time: 5.88 s
```

This was taking more than 60 seconds before making this change. 


Screen Shot 2020-04-08 at 11 38 12 AM

intake_esm/core.py
```diff
@@ -225,6 +229,13 @@ def __getitem__(self, key):
         else:
             raise KeyError(key)
 
+    def _get_entries(self):
+        """Eager evaluation of the delayed entries dictionary.
+           This is needed for enabling functionality provided by the base class
+           which expects a regular dictionary."""
+        entries = dask.compute(self._entries)[0]
+        return entries
```
jhamman, line 237: I was thinking of something a bit more lightweight (I don't think we want to involve dask at this point).

I'm not 100% sure how to do this but I'm thinking a just-in-time entry creation would work:

```python
def __getitem__(self, key):
    try:
        return self._entries[key]
    except KeyError:
        if key in self.keys():
            self._entries[key] = make_entry(key)
            return self._entries[key]
        else:
            raise KeyError
```
Do you think something like this could be made to work?

andersy005: > Do you think something like this could be made to work?

I think so. We may need to override a few methods from the base class. I will look into it, and will get back to you.

andersy005: @jhamman, should `__getitem(self, key)__` return an Entry instead of an `esm_datastore()` object?

andersy005: Per [intake](https://github.com/intake/intake/blob/cd3873e2933327e87e095e0ea653aacca72ab126/intake/catalog/base.py#L142), 

        * ``__iter__()`` -> an iterator of entry names
        * ``__getitem__(key)`` -> an Entry
        * ``items()`` -> an iterator of ``(key, Entry)`` pairs
        For best performance the object should also implement:
        * ``__len__()`` -> int
        * ``__contains__(key)`` -> boolean

andersy005: @jhamman, the latest commit address the just in time entry creation.

intake_esm/core.py
```diff
@@ -58,31 +57,171 @@ class esm_datastore(intake.catalog.Catalog):
...
+        return keys
+
+    @property
+    def key_template(self):
+        return self.sep.join(self.aggregation_info['groupby_attrs'])
+
+    def __len__(self):
+        return len(self.keys())
+
+    def _get_entries(self):
```
andersy005, line 176: This method is called by `self.items()`. So, I went ahead and made sure to load all entries missing from `self._entries`.

intake_esm/core.py
```diff
@@ -58,31 +57,171 @@ class esm_datastore(intake.catalog.Catalog):
...
+            return self._entries[key]
+        except KeyError:
+            if key in self.keys():
+                _key = tuple(key.split(self.sep))
+                df = self._grouped.get_group(_key)
+                self._entries[key] = _make_entry(key, df, self.aggregation_info)
+                return self._entries[key]
+            else:
+                raise KeyError(key)
+
```
andersy005, line 197: @jhamman, I like this implementation of `__getitem__()`. It clearly draws a line between searching for an item, and getting an item. Thank you for the idea!

intake_esm/core.py
```diff
@@ -58,31 +57,171 @@ class esm_datastore(intake.catalog.Catalog):
...
+    def key_template(self):
+        return self.sep.join(self.aggregation_info['groupby_attrs'])
+
+    def __len__(self):
+        return len(self.keys())
+
+    def _get_entries(self):
+        # Due to just-in-time entry creation, we may not have all entries loaded
+        # We need to make sure to create entries missing from self._entries
+        if len(self._entries) != len(self.keys()):
```
jhamman, line 179: I wonder if we could simplify this a bit:

```python
missing = set(self.keys()) - set(self._entries.keys())
for key in missing:
   _ = self[key]
```

intake_esm/core.py
```diff
@@ -351,142 +472,70 @@ def to_dataset_dict(
...
-
-        dsets = client.gather(futures)
-        self._ds = {group_id: ds for (group_id, ds) in dsets}
-
-        if _is_client_local:
-            client.close()
-            del client
-
-        return self._ds
+    to_xarray = to_dask = to_dataset_dict
```
jhamman, line 531: I would lean toward removing `to_xarray` and `to_dask` as methods on the catalog. Other Intake catalogs don't implement these methods AND they really belong on individual data sources.

intake_esm/source.py
```diff
@@ -0,0 +1,127 @@
...
+    def __init__(
+        self,
+        df,
+        aggregation_dict,
+        path_column,
+        variable_column,
+        data_format=None,
+        format_column=None,
+        cdf_kwargs={'chunks': {}},
+        zarr_kwargs={},
```
jhamman, line 25: Do you think we can combine these into a single keyword argument? As far as I understand, they would never be simultaneously utilized.

tests/test_core.py
```diff
@@ -77,33 +88,58 @@ def test_unique():
...
+def test_load_esmcol_remote(zarr_aws_cesmle_col):
+    assert isinstance(zarr_aws_cesmle_col.df, pd.DataFrame)
+
+
+def test_getitem():
+    key = 'CMIP.CNRM-CERFACS.CNRM-CM6-1.historical.Amon.gr'
+    col = intake.open_esm_datastore(cdf_col_sample_cmip6)
+    x = col[key]
+    assert isinstance(x, intake.catalog.local.LocalCatalogEntry)
+
```
jhamman, line 100: We could also use a few tests using these new catalog entries. We want to make sure we can pass various combinations of arguments to the open functions and the combined datasets are as expected.

jhamman: It would also be good to test the construction of an Entry outside the catalog.

intake_esm/utils.py
```diff
@@ -107,33 +107,14 @@ def _fetch_catalog(collection_data, esmcol_path):
...
-                    _is_client_local = True
-            else:
-                client = Client(processes=False)
-                _is_client_local = True
-            return client, _is_client_local
-    except Exception as exc:
-        raise exc
+@lru_cache(maxsize=None)
+def _load_csv(path, **kwargs):
+    return pd.read_csv(path, **kwargs)
```
jhamman, line 120: Is this related to this PR?

andersy005: > Is this related to this PR?

Nope :) I removed it

tests/test_core.py
andersy005: 👍

I added a few more tests

intake_esm/source.py
andersy005: > As far as I understand, they would never be simultaneously utilized.

For catalogs with non-homogeneous assets (netcdf, zarr), you would likely utilize both arguments at the same. So far, I have not seen an example of a non-homogeneous catalog. However, since the esm-collection-spec allows this non-homogeneity, I figured that the easiest thing to do was to separate these two arguments. 

Let me know if you have suggestions on how to combine these two, and still support the non-homogeneous catalogs.

jhamman: Gotcha. I didn't know this was possible. Let's just leave this for now.

jhamman:
> * [ ]  How should we handle `aggregate=False`?

We should cover this in a separate PR. 

My last comment on this PR is that the docs need to be updated. We need to get a few others (@kmpaul and @matt-long) to take a look over things and/or try it out.

andersy005:
> My last comment on this PR is that the docs need to be updated. We need to get a few others (@kmpaul and @matt-long) to take a look over things and/or try it out.

Gentle ping @kmpaul & @matt-long

andersy005:
> There are a lot of code changes, here. It it true that this PR only makes col[key].to_dask() equivalent to col.to_dataset_dict()[key]? 

- Adds a few basic dict-like methods (`keys()`, `__len__()`, `__getitem()__`, `__contains__()`)

- Moves assets aggregation logic into a standalone data source class (`ESMGroupDataSource`).

- Allow users to access datasets through the Entry class (via `__getitem()`) and the `to_dask()`
    ```python
    dset_dict = col.to_dataset_dict()
    ds = dset_dict['AerChemMIP.BCC.BCC-ESM1.ssp370.AERmon.gn']
    ```

    is equivalent to 

    ```python
    ds = col['AerChemMIP.BCC.BCC-ESM1.ssp370.AERmon.gn'].to_ask()
    ```

All these changes & additions were necessary in order to make `col[key].to_dask()` equivalent to `col.to_dataset_dict()[key]` (which was the main objective of this PR).

andersy005:
> The col[key].to_dask() step takes quite a while to run. Do you know why?

I don't know where the speed discrepancies between `to_dask()`, and `to_dataset_dict()` are coming from yet. Under the hood, `to_dataset_dict()` invokes `to_dask()`. I expected both methods to take relatively the same amount time. 

Screen Shot 2020-04-22 at 12 08 24 PM

Screen Shot 2020-04-22 at 12 08 34 PM

kmpaul:
That's very odd.  I might do some profiling.  Maybe simple `print(time.clock())` profiling.  Just to see where the time is being spent.

jhamman:
@andersy005 / @kmpaul - any blockers on merging this PR at this point?

kmpaul:
@jhamman: I guess I can't think of any blockers to merge.  It works and looks good to me.  As long as we merged this PR and then look at the performance problems in another PR, I'm fine with that.

andersy005:
>. As long as we merged this PR and then look at the performance problems in another PR, I'm fine with that.

Noted

Thank you @jhamman & @kmpaul! Glad to see this brought to completion! I am going to merge this soon.

jhamman:
Thanks @andersy005! This is was a big lift and will be a very useful change to this package, nice work!

andersy005:
Thank you for the pushing us in the right direction :)

kmpaul:
Yes, thanks to you, both!

PR #195, opened 2020-01-17
Add catalog_dict functionality

bonnland:
This pull request will be incremental in nature, in the spirit of team collaboration.


bonnland:
Hi @andersy005, this may be a good point to review.   Perhaps there is documentation to work on; are the docstrings used for that in this project?   Also, I am open to the idea of adding 'orient' as a parameter to serialize() if you prefer this idea.

bonnland:
Please feel free to point out anything that could be improved in what I have done so far.

bonnland:
@andersy005  I am not clear on why the read-the-docs build is failing.

andersy005:
@bonnland, thank you for working on this!

> I am not clear on why the read-the-docs build is failing.

Can you merge the master branch into  `add-catalog-dict` branch on your fork?

```bash
git pull upstream master
```

I am going to provide a detailed review around 2pm today

intake_esm/core.py
```diff
@@ -157,11 +162,19 @@ def serialize(self, name, directory=None):
...
+            collection_data['catalog_file'] = csv_file_name.as_posix()
+            print(f'Writing csv catalog to: {csv_file_name}')
+            self.df.to_csv(csv_file_name, compression='gzip', index=False)
+        else:
+            catalog_length = len(self.df)
+            print(f'Writing catalog with {catalog_length} entries into: {json_file_name}')
+            if catalog_length > 100:
+                print("  (call with catalog_type='file' to save catalog as a separate CSV file)")
+            collection_data['catalog_dict'] = self.df.to_dict(orient='records')
+
```
andersy005: Since `catalog_dict` and `catalog_file` are mutually exclusive, before finalizing the serialization process, we need to make sure that `catalog_data` doesn't contain both `catalog_dict` and `catalog_file`. Here's what I mean:

- If the original json file has `catalog_file`,  with this line `collection_data = self._col_data.copy()`, we end up getting `catalog_file` in `collection_data`. If the user then specifies `catalog_type='dict'`, we end up creating a new key `catalog_dict` and adding the catalog table to this. As a result, our json file now has both `catalog_dict` and `catalog_file`. 

One way to address this issue is to delete `catalog_dict`, `catalog_file` keys regardless of whether they are in the dictionary, 

```python
collection_data = self._col_data.copy()
for key in {'catalog_dict', 'catalog_file'}: 
    _ = catalog_data.pop(key, None)
```

After this, we can proceed to the next steps:

```python
collection_data['id'] = name
...
```


Example:                                     

```python
In [7]: catalog_data                                                                                               
Out[7]: {'catalog_dict': 'foo', 'other_key': 'bar'}

In [8]: for key in {'catalog_dict', 'catalog_file'}: 
   ...:     _ = catalog_data.pop(key, None) 
   ...:                                                                                                            

In [9]: catalog_data                                                                                               
Out[9]: {'other_key': 'bar'}
```

bonnland: These are good suggestions.   I will put these changes in shortly.

bonnland:
@andersy005  I tried putting the changes you suggested in a method, in the spirit of self-documenting the code.   Again, I am open to suggested improvements.

intake_esm/core.py
```diff
@@ -278,6 +292,12 @@ def _get_subset(self, **query):
         query_results = self.df.loc[condition]
         return query_results.reset_index(drop=True)
 
+    def _clear_old_catalog(self, catalog_data):
+        """ Remove any old references to the catalog."""
+        for key in {'catalog_dict', 'catalog_file'}:
+            _ = catalog_data.pop(key, None)
+        return catalog_data
```
andersy005: :+1: 

I have one suggestion:  this function doesn't need to be an instance method (i.e have self as a parameter) since we are not really using any extra information from the class. I propose that we remove the `self` parameter and make this function a  private/nested function in the `serialize()` method:

```python
def serialize(self, name, directory=None, catalog_type='dict'):

  def _clear_old_catalog(catalog_data):
        """ Remove any old references to the catalog."""
        for key in {'catalog_dict', 'catalog_file'}:
            _ = catalog_data.pop(key, None)
        return catalog_data

  collection_data = _clear_old_catalog(collection_data)
...
```

What do you think?

bonnland: @andersy005 I think it's a good idea.  Do private functions usually go at the top of the parent function?

andersy005: > Do private functions usually go at the top of the parent function?

It's not required. They can be put anywhere. The only requirement is that by the time you call the function, it must be defined/declared.

bonnland:
@andersy005  I tried merging my master branch into this development branch, in order to fix the read-the-docs build, but it seems that I did not succeed.  I find all of the fetch, pull, merge, push, commands still confusing, I guess.

intake_esm/core.py
```diff
@@ -157,11 +169,20 @@ def serialize(self, name, directory=None):
...
-        self.df.to_csv(csv_file_name, compression='gzip', index=False)
+        if catalog_type == 'file':
+            collection_data['catalog_file'] = csv_file_name.as_posix()
+            print(f'Writing csv catalog to: {csv_file_name}')
+            self.df.to_csv(csv_file_name, compression='gzip', index=False)
+        else:
+            catalog_length = len(self.df)
+            print(f'Writing catalog with {catalog_length} entries into: {json_file_name}')
+            if catalog_length > 100:
+                print("  (call with catalog_type='file' to save catalog as a separate CSV file)")
```
andersy005: Thinking more about the `100` threshold here,  I am now in favor of removing this suggestion. IMHO, I don't think that there's a better default here. So, instead of us guessing what this default is, we should let user decide what do. It might be more useful to move this suggestion to the docstring instead.  After the parameter section in the docstring, we can add a `Notes` section:

```python
""""
Notes
------
# Add suggestion here
"""
```

bonnland: @andersy005  I'm not sure if I should add a callout to every message I make here, but anyway...I have pushed the change you suggested.

andersy005: > I'm not sure if I should add a callout to every message I make here

You don't have to :) I get notified by GitHub.

intake_esm/core.py
```diff
@@ -157,11 +174,18 @@ def serialize(self, name, directory=None):
...
 
-        print(f'Writing csv catalog to: {csv_file_name}')
-        self.df.to_csv(csv_file_name, compression='gzip', index=False)
+        if catalog_type == 'file':
+            collection_data['catalog_file'] = csv_file_name.as_posix()
+            print(f'Writing csv catalog to: {csv_file_name}')
+            self.df.to_csv(csv_file_name, compression='gzip', index=False)
+        else:
+            catalog_length = len(self.df)
+            print(f'Writing catalog with {catalog_length} entries into: {json_file_name}')
```
andersy005: Should we be reporting the `catalog_length` regardless of the `catalog_type` specified by the user?

bonnland: Good idea.

andersy005:
The Readthedocs issue is related to a more recent pip release which seems to have been released with a glitch in it: https://github.com/pypa/pip/issues/7620

PR #205, opened 2020-03-08
Use dask cluster if available

andersy005:
Since `v2019.10.15` `intake-esm` started using `concurrent.futures` when loading dataset groups. One of the shortcomings of this approach is that intake-esm would **always** spawn its own `threadpool` in order to load groups of datasets concurrently. 

This PR changes this in the following ways:

- If a user has a dask cluster already running, use it instead of spawning our own threadpool.
- If there is no dask cluster, we default to dask's threads scheduler by spawning a threadpool with `total number of threads = total number of groups` 

Also, this PR removes our half-baked progressbar functionality and uses the functionality provided by `dask.diagnostics.ProgressBar` and `distributed.progress`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,8 @@ shared: &shared
     - run:
         name: Running Tests
         command: |
-          pytest -n 8 --junitxml=test-reports/junit.xml --cov=./ --verbose
+          pytest --junitxml=test-reports/junit.xml --cov=./ --verbose
+        no_output_timeout: 30m
     - run:
         name: Uploading code coverage report
         command: |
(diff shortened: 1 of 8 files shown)


andersy005:
```python
In [1]: import intake                                                                       

In [2]: col_url = "https://raw.githubusercontent.com/NCAR/intake-esm-datastore/master/catalo
   ...: gs/pangeo-cmip6.json"                                                               

In [3]: col = intake.open_esm_datastore(col_url)                                            

In [4]: cat = col.search(experiment_id=['historical', 'ssp585'], table_id='Oyr', variable_id
   ...: ='o2', grid_label='gn')                                                             

In [5]: dset_dict = cat.to_dataset_dict(zarr_kwargs={'consolidated': True, 'decode_times': F
   ...: alse})                                                                              

--> The keys in the returned dictionary of datasets are constructed as follows:
        'activity_id.institution_id.source_id.experiment_id.table_id.grid_label'
                
--> There are 17 group(s)
[########################################] | 100% Completed |  6.0s
```

Issue #202, opened 2020-03-06
Select only datasets that fulfill *ALL* query criteria.


jbusecke:
I find myself using a workflow similar to this example very often:

- Define a set of experiments(`experiment_id=['piControl', 'historical'])

- Query the catalog and end up with a bunch of datasets, which fulfill *any* of the criteria, e.g. [ModelA.piControl, ModelB.historical, ModelC.piControl, ModelC.historical]

- Many analyses require to have both experiments, and so I end up writing a lot of logic to isolate those models and query the catalog again. 

Is there a way in intake-esm to force a combination of input criteria?

E.g. in the example above, if I could do: 
```
col.search(experiment_id=['piControl', 'historical'], force_all=True)
```
or similar, and I would get back only the datasets that have both experiments ([ModelC.piControl, ModelC.historical]), that would be amazing.

andersy005:
Thank you for opening this issue, @jbusecke! In the past, I've seen people using some other workarounds such as manipulating the dataframe itself. Supporting this would be useful for other users as well. 

After a quick glance, I do believe that flipping `|` to `&` on line (296) below is one way to address this issue:

```
condition_i = condition_i | (self.df[key] == val_i)
```
https://github.com/NCAR/intake-esm/blob/318a49176d485525f375e3a244740c152fe0f7ab/intake_esm/core.py#L288-L300

Can you post here one of the queries(you've had to manipulate) that I can use for testing purposes?

andersy005:
> After a quick glance, I do believe that flipping | to & on line (296) below may be one way to address this issue

I take this back 😀. I am currently looking into this, and will report back when I find something.

jbusecke:
> Can you post here one of the queries(you've had to manipulate) that I can use for testing purposes?

I have previously searched for all models that have both `variable_id=['uo', 'o2']`and all `experiment_id=['historical', 'piControl', 'ssp245' ]'. 

These models should have all of these fields available (['CNRM-ESM2-1', 'ACCESS-ESM1-5', 'MIROC-ES2L', 'GFDL-CM4']). I think `CanESM` and one of the `IPSL` ones might have them too but they caused different problems, so I eliminated them.

andersy005:
Thank you for the example queries. 

I was able to come up with a minimal example below:

```python
In [56]: import pandas as pd                                                                                       

In [57]: df = pd.DataFrame({'A': ["NCAR", "CSIRO", "NCAR", "IPSL", "IPSL"],  
[...]

PR #207, opened 2020-03-10
Enhanced search: enforce query criteria via `require_all_on` argument

andersy005:
Towards #202 

Cc @jbusecke, @matt-long

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -255,3 +255,9 @@ def test_get_subset():
 
     assert set(x.keys()) == set(['CESM', 'FOO'])
     assert list(x.values()) == [2, 2]
+
+    x = _get_subset(df, C=['control', 'hist'], D=['NO2'], force_all_on='B')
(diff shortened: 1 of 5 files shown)


tests/test_core.py
```diff
@@ -255,3 +255,9 @@ def test_get_subset():
 
     assert set(x.keys()) == set(['CESM', 'FOO'])
     assert list(x.values()) == [2, 2]
+
+    x = _get_subset(df, C=['control', 'hist'], D=['NO2'], force_all_on='B')
```
matt-long: Would `require_all_on` be a better name than `force_all_on`? I think it would.

andersy005: Done!

andersy005:
I am planning on merging this later this afternoon unless there is an objection and/or additional feedback.

jbusecke:
Testing right now. Will report back in a bit.

PR #209, opened 2020-03-11
Fix bug in _get_subset() when using `require_all_on`

andersy005:
This PR fixes the half-baked #207. See https://github.com/NCAR/intake-esm/issues/202#issuecomment-597325056 and https://github.com/NCAR/intake-esm/issues/202#issuecomment-597335795

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -259,5 +259,18 @@ def test_get_subset():
     x = _get_subset(df, C=['control', 'hist'], D=['NO2'], require_all_on='B')
     assert x.empty
 
+    x = _get_subset(df, C=['control', 'hist'], D=['O2', 'NO2'], require_all_on='B')
+    assert x.empty
+
     x = _get_subset(df)
     assert x.empty
+
+    x = _get_subset(df, C=['control', 'hist'], D=['O2'], require_all_on='B')
+    expected = [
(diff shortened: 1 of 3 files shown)


tests/test_core.py
```diff
@@ -259,5 +259,18 @@ def test_get_subset():
     x = _get_subset(df, C=['control', 'hist'], D=['NO2'], require_all_on='B')
     assert x.empty
 
+    x = _get_subset(df, C=['control', 'hist'], D=['O2', 'NO2'], require_all_on='B')
+    assert x.empty
+
```
jbusecke: Do you think it would be worth adding 
```
_get_subset(df, C=['hist'], D=['NO2', 'O2'], require_all_on='B')

expected = [
         {'A': 'IPSL', 'B': 'FOO', 'C': 'hist, 'D': 'NO2'},
         {'A': 'IPSL', 'B': 'FOO', 'C': 'hist', 'D': 'O2'},
     ]
```
just to make sure this works on different columns?

andersy005: Good catch. I added a few more tests including this one.

andersy005:
@jbusecke, once you give a green light to this I will go ahead and merge it.

jbusecke:
Looks great. Go for it!

PR #229, opened 2020-04-30
Add html representation for the catalog object

andersy005:
This provides more information about the catalog object as an HTML repr. 

Screen Shot 2020-04-30 at 1 05 31 PM


andersy005:
Thank you, @bonnland! Unless there is any objection, I am going to merge this soon.

bonnland:
It looks pretty non-controversial to me.

PR #235, opened 2020-05-08
Provide informative message/warnings from empty queries

### LINKED ISSUE #234: informative message from empty queries
I tried the `to_dataset_dict` method on a query that happened to be empty. This error message led me astray.

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-13-fd498b130342> in <module>
----> 1 dsets = catalog.search(experiment=experiment, variable=variable).to_dataset_dict(cdf_kwargs=cdf_kwargs)
      2 dsets

/glade/work/mclong/miniconda3/envs/cesm2-marbl/lib/python3.7/site-packages/intake_esm/core.py in to_dataset_dict(self, zarr_kwargs, cdf_kwargs, preprocess, storage_options, progressbar)
    593                 )
    594 
--> 595             with concurrent.futures.ThreadPoolExecutor(max_workers=len(sources)) as executor:
    596                 future_tasks = [executor.submit(_load_source, source) for source in sources]
    597 

/glade/work/mclong/miniconda3/envs/cesm2-marbl/lib/python3.7/concurrent/futures/thread.py in __init__(self, max_workers, thread_name_prefix, initializer, initargs)
    127             max_workers = (os.cpu_count() or 1) * 5
    128         if max_workers <= 0:
--> 129             raise ValueError("max_workers must be greater than 0")
    130 
    131         if initializer is not None and not callable(initializer):

ValueError: max_workers must be greater than 0
```

I am using 'v2020.5.1'

andersy005:
Checklist

- [x] Tests added / passed

### DIFF
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -54,7 +54,7 @@ def test_esm_group(df, aggregation_dict):
 
 
 def test_esm_group_empty_df(df, aggregation_dict):
-    empty_df = _get_subset(df)
+    empty_df = pd.DataFrame(columns=df.columns)
     args = dict(
         df=empty_df,
         aggregation_dict=aggregation_dict,
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -51,3 +51,19 @@ def test_esm_group(df, aggregation_dict):
     assert set(subset_df['member_id']) == set(ds['member_id'].values)
     source.close()
     assert source._ds is None
+
+
+def test_esm_group_empty_df(df, aggregation_dict):
+    empty_df = _get_subset(df)
+    args = dict(
+        df=empty_df,
+        aggregation_dict=aggregation_dict,
+        path_column='path',
+        variable_column='variable_id',
+        data_format='netcdf',
+        format_column=None,
+        cdf_kwargs={'chunks': {'time': 2}},
+    )
+
+    with pytest.raises(ValueError, match=r'`df` must be a non-empty pandas.DataFrame'):
+        _ = ESMGroupDataSource(**args)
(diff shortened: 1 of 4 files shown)


tests/test_source.py
```diff
@@ -51,3 +51,19 @@ def test_esm_group(df, aggregation_dict):
     assert set(subset_df['member_id']) == set(ds['member_id'].values)
     source.close()
     assert source._ds is None
+
+
+def test_esm_group_empty_df(df, aggregation_dict):
+    empty_df = _get_subset(df)
```
bonnland, line 57: Not asking to change anything, but this line is confusing to me.  Maybe I haven't seen this type of test structure before.   I normally don't see a dataframe getting passed into a test, so I don't understand what value it has, hence why to expect it should be empty.

andersy005: Okay, I see how this may be confusing 

> I normally don't see a dataframe getting passed into a test, so I don't understand what value it has, hence why to expect it should be empty.

By default, `_get_subset()` returns an empty dataframe when the query is empty. So, calling `_get_subset(df)` is going to return an **empty** dataframe:
 
https://github.com/NCAR/intake-esm/blob/b906b7b3c7ba3e6fce6b0d15fa1b2cf1d5f9eec3/intake_esm/core.py#L623-L626

andersy005: I updated the test to use a more explicit option i.e. I manually created the empty dataframe.

andersy005:
Thank you @matt-long & @bonnland for your feedback! I am going to merge this.

PR #236, opened 2020-05-09
Support regular expression objects in `search()`

### LINKED ISSUE #92: Extend search method to filter rows in collection by wildcard
It would be nice to extend our search method to allow passing in regular expressions. 

The current search method works for **exact matches** only. 

For example, the following works out of the box:

```python
col.search(variable=['TEMP'], case=['b.e11.BRCP85C5CNBDRD.f09_g16.103', 'b.e11.BRCP85C5CNBDRD.f09_g16.102'])
```

With regular expression, you could search for something like:

```python
col.search(variable=['TEMP'], case='b.e11.B20TRC5CNBDRD.f09_g16.10*')

```

Ccing

@matt-long, @kmpaul, @jukent

andersy005:
- [x] Waiting on #235 

- Fixes https://github.com/NCAR/cesm-lens-aws/issues/48

- Fixes #92

## Examples

```python
In [1]: import intake                                                                                  

In [2]: col = intake.open_esm_datastore("https://raw.githubusercontent.com/NCAR/cesm-lens-aws/master/in
   ...: take-catalogs/aws-cesm1-le.json")                                                              

In [3]: import re                                                                                      

In [18]: variable_patterns = [re.compile('VEL$'), re.compile('TAU*')]                                  

In [19]: col.search(variable=variable_patterns).df.variable                                            
Out[19]: 
0      TAUX
1     TAUX2
2      TAUY
3     TAUY2
4      UVEL
5      VVEL
6      WVEL
7      TAUX
8     TAUX2
9      TAUY
10    TAUY2
11     UVEL
12     VVEL
13     WVEL
14     TAUX
15    TAUX2
16     TAUY
17    TAUY2
18     UVEL
19     VVEL
20     WVEL
```

Checklist

- [x] Tests added / passed

Cc @jeffdlb

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -646,26 +646,6 @@ def _unique(df, columns=None):
 
 
 def _get_subset(df, require_all_on=None, **query):
-    """
-    Utility search function.
-
-    Parameters
-    ----------
-    df : pandas.DataFrame
-        A pandas dataframe to perform a query on.
-    require_all_on : list, str, optional
-        A dataframe column or a list of dataframe columns across
-        which all entries must satisfy the query criteria.
-        If None, return entries that fulfill any of the criteria specified
-        in the query, by default None.
-    **query:
-        keyword arguments corresponding to user's query to execute against the dataframe.
-
-    Returns
-    -------
-    pandas.DataFrame
-        A pandas dataframe containing found results.
-    """
     message = 'Query returned zero results.'
     if not query:
         warn(message)
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -1,6 +1,7 @@
 import itertools
 import json
 import logging
+import re
 from collections.abc import Iterable
 from warnings import warn
 
@@ -303,15 +304,19 @@ class esm_datastore(intake.catalog.Catalog):
         )
 
     def search(self, require_all_on=None, **query):
-        """Search for entries in the catalog.
+        """
+        Search for entries in the catalog. This method also accepts compiled
+        regular expression objects from :py:method:`~re.compile` as patterns.
 
         Parameters
         ----------
-        require_all_on : str, list
-           name of columns to use when enforcing the query criteria.
-           For example: `col.search(experiment_id=['piControl', 'historical'], require_all_on='source_id')`
-           returns all assets with `source_id` that has both `piControl` and `historical`
-           experiments.
+        require_all_on : list, str, optional
+            A dataframe column or a list of dataframe columns across
+            which all entries must satisfy the query criteria.
+            If None, return entries that fulfill any of the criteria specified
+            in the query, by default None.
+        **query:
+            keyword arguments corresponding to user's query to execute against the dataframe.
 
         Returns
         -------
@@ -336,6 +341,20 @@ class esm_datastore(intake.catalog.Catalog):
         260        CMIP            BCC  BCC-CSM2-MR  ...         gn  gs://cmip6/CMIP/BCC/BCC-CSM2-MR/historical/r1i...            NaN
         346        CMIP            BCC  BCC-CSM2-MR  ...         gn  gs://cmip6/CMIP/BCC/BCC-CSM2-MR/historical/r2i...            NaN
         401        CMIP            BCC  BCC-CSM2-MR  ...         gn  gs://cmip6/CMIP/BCC/BCC-CSM2-MR/historical/r3i...            NaN
+
+        The search method also accepts compiled regular expression objects
+        from :py:method:`~re.compile` as patterns.
+
+        >>> import re
+        >>> # Let's search for variables containing "Frac" in their name
+        >>> pat = re.compile(r'Frac') # Define a regular expression
+        >>> cat.search(variable_id=pat)
+        >>> cat.df.head().variable_id
+        0     residualFrac
+        1    landCoverFrac
+        2    landCoverFrac
+        3     residualFrac
+        4    landCoverFrac
         """
 
         results = _get_subset(self.df, require_all_on=require_all_on, **query)
@@ -627,6 +646,26 @@ def _unique(df, columns=None):
 
 
 def _get_subset(df, require_all_on=None, **query):
+    """
+    Utility search function.
+
+    Parameters
+    ----------
+    df : pandas.DataFrame
+        A pandas dataframe to perform a query on.
+    require_all_on : list, str, optional
+        A dataframe column or a list of dataframe columns across
+        which all entries must satisfy the query criteria.
+        If None, return entries that fulfill any of the criteria specified
+        in the query, by default None.
+    **query:
+        keyword arguments corresponding to user's query to execute against the dataframe.
+
+    Returns
+    -------
+    pandas.DataFrame
+        A pandas dataframe containing found results.
+    """
     message = 'Query returned zero results.'
     if not query:
         warn(message)
@@ -635,13 +674,16 @@ def _get_subset(df, require_all_on=None, **query):
 
     query = _normalize_query(query)
     for key, val in query.items():
-        if isinstance(val, (tuple, list)):
-            condition_i = np.zeros(len(df), dtype=bool)
-            for val_i in val:
+        condition_i = np.zeros(len(df), dtype=bool)
+        for val_i in val:
+            if isinstance(df[key].dtype, (np.object, pd.core.arrays.string_.StringDtype)):
+                if isinstance(val_i, re.Pattern):
+                    condition_i = condition_i | (df[key].str.contains(val_i, regex=True))
+                else:
+                    condition_i = condition_i | (df[key] == val_i)
+            else:
                 condition_i = condition_i | (df[key] == val_i)
-            condition = condition & condition_i
-        elif val is not None:
-            condition = condition & (df[key] == val)
+        condition = condition & condition_i
     query_results = df.loc[condition]
 
     if require_all_on:
@@ -686,7 +728,7 @@ def _get_subset(df, require_all_on=None, **query):
 def _normalize_query(query):
     q = query.copy()
     for key, val in q.items():
-        if isinstance(val, str):
+        if isinstance(val, str) or not isinstance(val, Iterable):
             q[key] = [val]
     return q
 
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -2,14 +2,15 @@ import itertools
 import json
 import logging
 from collections.abc import Iterable
-from typing import Pattern
+from typing import Any, Dict, List, Pattern, Union
 from warnings import warn
 
 import dask
 import intake
 import numpy as np
 import pandas as pd
-from tqdm import tqdm
+import xarray as xr
+from fastprogress.fastprogress import progress_bar
 
 from .utils import _fetch_and_parse_json, _fetch_catalog, logger
 
@@ -71,11 +72,11 @@ class esm_datastore(intake.catalog.Catalog):
 
     def __init__(
         self,
-        esmcol_obj,
-        esmcol_data=None,
-        progressbar=True,
-        sep='.',
-        log_level='CRITICAL',
+        esmcol_obj: Union[str, pd.DataFrame],
+        esmcol_data: Dict[str, Any] = None,
+        progressbar: bool = True,
+        sep: str = '.',
+        log_level: str = 'CRITICAL',
         **kwargs,
     ):
 
@@ -89,14 +90,15 @@ class esm_datastore(intake.catalog.Catalog):
 
         if isinstance(esmcol_obj, str):
             self.esmcol_data, self.esmcol_path = _fetch_and_parse_json(esmcol_obj)
-            self.df = _fetch_catalog(self.esmcol_data, esmcol_obj)
+            self._df, self.catalog_file = _fetch_catalog(self.esmcol_data, esmcol_obj)
 
         elif isinstance(esmcol_obj, pd.DataFrame):
             if esmcol_data is None:
-                raise ValueError(f"Missing required argument: 'esmcol_data'")
-            self.df = esmcol_obj
+                raise ValueError("Missing required argument: 'esmcol_data'")
+            self._df = esmcol_obj
             self.esmcol_data = esmcol_data
             self.esmcol_path = None
+            self.catalog_file = None
         else:
             raise ValueError(f'{self.name} constructor not properly called!')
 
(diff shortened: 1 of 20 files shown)


andersy005:
> I've looked at the conversations around this issue, and I'm not sure that making the user compile regular expressions is the ideal approach. 

Partial string matches constitute a special case in my opinion. Using compiled regular expressions allow us to support other cases other than substring matches. Without compiled regular expressions , we would need to do a lot of work in intake-esm due to how pattern, regular expression matches/searches are done in Pandas: https://pandas.pydata.org/docs/reference/api/pandas.Series.str.contains.html#pandas.Series.str.contains.

bonnland:
> > I've looked at the conversations around this issue, and I'm not sure that making the user compile regular expressions is the ideal approach.
> 
> Partial string matches constitute a special case in my opinion. **Should we really support a special case if we can support a more general case?**
> 
Sometimes, yes.   Especially if it gets used more by users. :)  But I see your point of view, and I'm not saying it is wrong.   Did Pandas make a decision against supporting glob patterns for some reason?  That would surprise me.

andersy005:
> Did Pandas make a decision against supporting glob patterns for some reason?

AFAIK, pandas doesn't restrict the user from using glob patterns. The user just needs to provide additional parameters to pandas string methods instructing Pandas on how it should treat the search pattern: https://pandas.pydata.org/docs/reference/api/pandas.Series.str.contains.html#pandas.Series.str.contains

bonnland:
>  The user just needs to provide additional parameters to pandas string methods instructing Pandas on how it should treat the search pattern

A nice idea; I think users would be happy with a similar pattern for intake-esm, like a "contains" method.   Perhaps there is room for both.

andersy005:
> A nice idea; I think users would be happy with a similar pattern for intake-esm, like a "contains" method. Perhaps there is room for both.

As I pointed out, this might require a huge amount of work on our end. How can we do this with as little work as possible? I am open to suggestions on this.

andersy005:
> I've looked at the conversations around this issue, and I'm not sure that making the user compile regular expressions is the ideal approach.

Is 
```python
import re 
col.search(variable=re.compile('TAU*'))
```
a big ask? 

With glob patterns this corresponds to

```python
col.search(variable='TAU*')
```

but this requires additional branching logic in intake-esm depending on which metacharacters used in the expression.

bonnland:
>  How can we do this with as little work as possible? I am open to suggestions on this.
My original thinking is probably wrong, but I had the impression that because the heart of the intake-esm is a Pandas catalog, we could pass the work onto Pandas.   Perhaps I need to see the implementation more closely now.

bonnland:
> ```python
> col.search(variable=re.compile('TAU*'))
> ```

If this means "Any string following 'TAU'", then it is not a big ask.   If it means "TA" followed by any number of U's, then it is bigger ask for sure.

andersy005:
> If it means "TA" followed by any number of U's, then it is bigger ask for sure.

This is the decision I am trying to outsource to the user i.e. Let the user define whatever they want (https://docs.python.org/3/library/re.html) without us having to come up our own interpretation of what they want.

bonnland:
Your example was good, in the sense that you passed in something you thought would work, but the re syntax was not what you thought.   I think you are right in supporting these kinds of searches.   I think it would be nice to have a "helper" method that compiles a string into the correct regular expression for partial word matches.   In your example, this is `re.compile('TAU.*')`.   It is very easy to forget the dot, or what it means.   That was my point.

jeffdlb:
Thanks for working on this! I agree that full regexp support is better if feasible.

andersy005:
@jeffdlb has an [example notebook](https://github.com/NCAR/cesm-lens-aws/blob/master/notebooks/EnhancedIntakeCatalogDemo.ipynb) in which he tries to use substring matches on an enhanced catalog. With this PR, the user should be able to achieve the same thing, but with minimal amount of code:

```python
In [1]: url = 'https://ncar-cesm-lens.s3-us-west-2.amazonaws.com/catalogs/aws-cesm1-le-enhanced.js
   ...: on'                                                                                       

In [2]: import intake                                                                             

In [3]: col = intake.open_esm_datastore(url)     

In [10]: cat = col.search(long_name=re.compile('Wind.*'))                                                                  

In [11]: cat                                                                                                               
Out[11]: <aws-cesm1-le catalog with 3 dataset(s) from 12 asset(s)>

In [12]: cat.df.head()                                                                                                     
Out[12]: 
  component dim  ...                                  long_name                                               path
0       ocn  2D  ...            Wind stress in grid-x direction  s3://ncar-cesm-lens/ocn/monthly/cesmLE-20C-TAU...
1       ocn  2D  ...            Wind stress in grid-x direction  s3://ncar-cesm-lens/ocn/monthly/cesmLE-RCP85-T...
2       ocn  2D  ...            Wind stress in grid-x direction  s3://ncar-cesm-lens/ocn/monthly/cesmLE-CTRL-TA...
3       ocn  2D  ...  Wind stress (squared) in grid-x direction  s3://ncar-cesm-lens/ocn/monthly/cesmLE-20C-TAU...
4       ocn  2D  ...  Wind stress (squared) in grid-x direction  s3://ncar-cesm-lens/ocn/monthly/cesmLE-RCP85-T...

[5 rows x 9 columns]

In [13]: cat.nunique()                                                                                                     
Out[13]: 
component      1
dim            1
frequency      1
experiment     3
start          3
end            3
variable       4
long_name      4
path          12
dtype: int64

In [14]: cat.unique('long_name')                                                                                           
Out[14]: 
{'long_name': {'count': 4,
  'values': ['Wind stress (squared) in grid-x direction',
   'Wind stress (squared) in grid-y direction',
   'Wind stress in grid-x direction',
   'Wind stress in grid-y direction']}}

In [15]: cat.unique('variable')                                                                                            
Out[15]: {'variable': {'count': 4, 'values': ['TAUX', 'TAUX2', 'TAUY', 'TAUY2']}}
```

I could use more feedback on this PR i.e. Did I miss anything?, is the end user facing API good enough?

If people deem this PR good to go, please let me know

sherimickelson:
Is this approach case sensitive?  Will the re.IGNORECASE flag work if the user adds it?  For example, 
```
re.compile('Wind.*',re.IGNORECASE)
```

kmpaul:
@sherimickelson Yes.  That pattern should work.

intake_esm/core.py
```diff
@@ -660,13 +676,16 @@ def _get_subset(df, require_all_on=None, **query):
...
+                if isinstance(val_i, Pattern):
+                    condition_i = condition_i | (df[key].str.contains(val_i, regex=True))
+                else:
+                    condition_i = condition_i | (df[key] == val_i)
+            else:
                 condition_i = condition_i | (df[key] == val_i)
-            condition = condition & condition_i
-        elif val is not None:
-            condition = condition & (df[key] == val)
+        condition = condition & condition_i
```
kmpaul, line 688: @andersy005 You are doing a lot of `isinstance` checking here.  Do you need it?  Is seems that the `search()` method is just a "pass-through" interface to the `pandas.DataFrame[col].str.contains()` method.  If so, why not let Pandas do the error checking for you?  You could really simplify the code a lot and just tell people that:

```python
esm_datastore.search(key=val)
```

is equivalent to:

```python
esm_datastore.df[key].str.contains(val)
```

Are there any other parts of the API that are thin layers on top of Pandas?  If so, then we can pass off the error checking to Pandas.

andersy005: >  It seems that the search() method is just a "pass-through" interface to the pandas.DataFrame[col].str.contains() method

This would only be true if you had a pandas dataframe in which all column dtypes are strings. However, in most cases, the pandas dataframe has columns with a mix of dtypes. For instance, when  a user calls `esm_datastore.search(key1=val1, key2=val2)`, the keys here (`key1`, `key2` may refer to columns with different dtypes, and the `.str` accessor is available for string-like dtypes, and this is why I am doing the type checking.

andersy005: > If so, why not let Pandas do the error checking for you? 

Could you expand on this? I am asking because if you assume that all columns are of `string` dtype, you get the following error, and I don't see how intake-esm should handle this. 

```python
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({'key1' :[1, 2], 'key2': ['foo', 'bar']})

In [3]: df
Out[3]:
   key1 key2
0     1  foo
1     2  bar

In [4]: df.key1.str
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-4-ba969f2734e5> in <module>
----> 1 df.key1.str

~/opt/miniconda3/envs/analysis/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5268             or name in self._accessors
   5269         ):
-> 5270             return object.__getattribute__(self, name)
   5271         else:
   5272             if self._info_axis._can_hold_identifiers_and_holds_name(name):

~/opt/miniconda3/envs/analysis/lib/python3.7/site-packages/pandas/core/accessor.py in __get__(self, obj, cls)
    185             # we're accessing the attribute of the class, i.e., Dataset.geo
    186             return self._accessor
--> 187         accessor_obj = self._accessor(obj)
    188         # Replace the property with the accessor object. Inspired by:
    189         # http://www.pydanny.com/cached-property.html

~/opt/miniconda3/envs/analysis/lib/python3.7/site-packages/pandas/core/strings.py in __init__(self, data)
   2039
   2040     def __init__(self, data):
-> 2041         self._inferred_dtype = self._validate(data)
   2042         self._is_categorical = is_categorical_dtype(data)
   2043         self._is_string = data.dtype.name == "string"

~/opt/miniconda3/envs/analysis/lib/python3.7/site-packages/pandas/core/strings.py in _validate(data)
   2096
   2097         if inferred_dtype not in allowed_types:
-> 2098             raise AttributeError("Can only use .str accessor with string values!")
   2099         return inferred_dtype
   2100

AttributeError: Can only use .str accessor with string values!
```

andersy005:
> However, I might ask you consider how much of the API is really just a thin layer on top of the API to pandas.DataFrame

I think that the current API isn't just a thin layer on top of Pandas because we may or may not have a pandas dataframe with homogeneous dtypes. So, due to this heterogeneous nature of column data types, we need to apply different logic to different columns depending on what their data types are. For example, in the future we may need to support querying columns with datetime objects or geometry objects, and this would require having special logic for columns with these kinds of data types.

intake_esm/core.py
kmpaul: You are correct that this `search` method only maps to `.str.contains` if the `dtype` of the column is `string`, so my comment doesn't make sense.

My mind is just trying to figure out how to reduce the number of `if`-checks because there is only 1 case where you want to use `str.contains`, and all other cases are just `==` operations.  Maybe there isn't a way.

I guess you can mark this as resolved.

mnlevy1981: > My mind is just trying to figure out how to reduce the number of `if`-checks because there is only 1 case where you want to use `str.contains`

you could do something like

```python
        for val_i in val:
            if isinstance(df[key].dtype, (np.object, pd.core.arrays.string_.StringDtype)) and isinstance(val_i, Pattern):
                new_condition = (df[key].str.contains(val_i, regex=True))
            else:
                new_condition = (df[key] == val_i)
            condition_i = condition_i | new_condition
```

kmpaul: ...Let's see.  You have 2 conditions that must be met to use `str.contains`, and if either of those 2 conditions are not met, then you just use `==`, correct?

So, you could improve readability of the code by avoiding the nested `if` block and doing something like:

```python
column_is_stringtype = isinstance(df[key].dtype, (np.object, pd.core.arrays.string_.StringDtype))
value_is_repattern = isinstance(val_i, Pattern)

if column_is_stringtype and value_is_repattern:
    cond = df[key].str.contains(val_i, regex=True)
else:
    cond = (df[key] == val_i)

condition_i = condition_i | cond
```

kmpaul:
@andersy005 Fair enough.  Aesthetically, though, I would prefer to avoid nested `if`-blocks.  I suggested an alternative above.  Let me know what you think.

kmpaul:
...And actually, this is more than aesthetic.  Nested `if`-blocks make the code less readable.

andersy005:
> ...And actually, this is more than aesthetic. Nested `if`-blocks make the code less readable.

I concur. I will simplify the logic using your suggestion from https://github.com/NCAR/intake-esm/pull/236#discussion_r441724918

tests/test_core.py
```diff
@@ -391,17 +392,38 @@ def test_read_catalog_dict():
...
+            {'A': 'NCAR', 'B': 'CESM', 'C': 'hist', 'D': 'O2'},
+            {'A': 'CSIRO', 'B': 'BAR', 'C': 'control', 'D': 'O2'},
+            {'A': 'NCAR', 'B': 'CESM', 'C': 'control', 'D': 'O2'},
+        ],
+    ),
+    ({'C': ['hist'], 'D': ['TA']}, None, [{'A': 'NCAR', 'B': 'WACM', 'C': 'hist', 'D': 'TA'}],),
+    (
+        {
+            'C': [re.compile('hist.*', flags=re.IGNORECASE)],
+            'D': [re.compile('TA.*', flags=re.IGNORECASE)],
```
andersy005, line 408: @sherimickelson, I was able to test that the search function is respecting regex flags such as case sensitivity.

jeffdlb:
> ...And actually, this is more than aesthetic. Nested `if`-blocks make the code less readable.

How about a venerable Fortran 77 arithmetic IF or computed GO TO? :)

andersy005:
I plan to merge this today unless there is any objection

kmpaul:
None from me.

jeffdlb:
Thank you so much for working on this, Anderson!

markusritschel:
How do you think about a compromise or rather a combined solution? I think that a wildcard approach (i.e. `TAU*` = 'TAU' followed by any string) allowed in the direct input as well as, for more complex queries, a compiled regex (which also gives all the options of `re`, e.g. IGNORECASE etc.) would be a nice solution. What's your opinion?

Thanks for your work on this! This makes queries a lot more flexible.

markusritschel:
Is this update already in the conda repo available?
I just updated `intake-esm` but cannot get the regex support working. When I hand over a compiled regex to the search method I get a 'catalog with 0 dataset(s) from 0 asset(s)'.

andersy005:
> How do you think about a compromise or rather a combined solution? I think that a wildcard approach (i.e. `TAU*` = 'TAU' followed by any string) allowed in the direct input as well as, for more complex queries, a compiled regex (which also gives all the options of `re`, e.g. IGNORECASE etc.) would be a nice solution. What's your opinion?

👍 

@markusritschel, I am going to look into the wildcard approach


> Is this update already in the conda repo available?

Not yet. You will need to use the master branch for the time being:

```bash
python -m pip install git+https://github.com/intake/intake-esm.git@master
```

markusritschel:
Ok, thanks for the hint :-) :+1:

Issue #231, opened 2020-05-04
Progressbar gets chopped up by preprocessing print statements.


jbusecke:
I just played a little bit with the newest version and noticed that print statements from preprocessing end up chopping the status bar into this super long thing: 
![image](https://user-images.githubusercontent.com/14314623/81019407-9b1c7b00-8e34-11ea-8429-8e9489460c7b.png)

Is there a way to prevent that? Happy to implement in cmip6_pp upstream if that would help.

andersy005:
This is a tqdm issue: https://github.com/tqdm/tqdm/issues/375. There is a workaround for this: https://github.com/tqdm/tqdm/issues/375#issuecomment-328466265. https://github.com/NCAR/intake-esm/blob/31ba8b5818c2f8255c24be52463302bd9b2062e4/intake_esm/core.py#L10

Would you be interested in submitting a PR for this? 

Another temporary solution is to disable the progressbar (assuming you don't need it :)).

jbusecke:
Is it literally just that one line I have to replace? Then, of course.

andersy005:
> Is it literally just that one line I have to replace? 

I think so. Try testing the one line change with your existing notebook to see if the problem goes away. Looking forward to the PR :)

andersy005:
Should be fixed by #238

jbusecke:
Just updated to the newest master on my working env, and I just wanted to say: I really like the look of the new progressbars!

andersy005:
Glad to hear that the new progress-bar is much better! Closing....

PR #237, opened 2020-05-12
switched tqdm to auto

jbusecke:
towards #231 

I tested it in a local notebook and it seems like this did the trick. It shows a widget progressbar and the warnings are lined up underneath.

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -8,7 +8,7 @@ import dask
 import intake
 import numpy as np
 import pandas as pd
-from tqdm import tqdm
+from tqdm.auto import tqdm
 
 from .utils import _fetch_and_parse_json, _fetch_catalog, logger
 


jbusecke:
Not sure what I messed up in the formatting. The flake8 message does not provide very useful feedback.

jbusecke:
On second view, it is a bit ugly : 
image
Any idea why this doesnt show up like a nice progressbar?

jbusecke:
It might be that the extension is not installed? E.g. see [here](https://github.com/tqdm/tqdm/issues/394#issuecomment-624531391).

Either way this is already a big improvement for my workflow.

intake_esm/core.py
```diff
@@ -8,7 +8,7 @@
 import intake
 import numpy as np
 import pandas as pd
-from tqdm import tqdm
+from tqdm.auto import tqdm
```
andersy005, line 11: How about wrapping the imports in a try block so that when `ipywidgets` is not installed, we revert back to the default `tqdm`

```python
try:
    import ipywidgets
    from tqdm.auto import tqdm
except ImportError as e:
    from tqdm import tqdm
```

andersy005:
> On second view, it is a bit ugly :
> image
> Any idea why this doesnt show up like a nice progressbar? Even having it in only one line would be nicer.

I don't know what's going on here :(. I've run into issues with tqdm before, and I always ended up giving up :). I've been looking at fastprogress: https://github.com/fastai/fastprogress, and it appears to be an alternative to tqdm. I will look into it, and will let you know whether it might be better to just replace tqdm with fastprogress.

andersy005:
> I've been looking at fastprogress: fastai/fastprogress, and it appears to be an alternative to tqdm. I will look into it, and will let you know whether it might be better to just replace tqdm with fastprogress.


@jbusecke, I made a few changes in #238. When you get a chance please take a look at this [notebook](https://nbviewer.jupyter.org/github/andersy005/intake-esm/blob/53b0a8be8b35ba20cdf9d4ad5757989ff723a8d5/docs/source/notebooks/tutorial.ipynb) and let me know what you think.

andersy005:
Superseded by #238

PR #238, opened 2020-05-12
Use fastprogress

andersy005:
Checklist

- [ ] Tests added / passed

- supersedes #237


jbusecke:
Just gave it a quick go and got some mixed results. The progressbar stays in one piece, but appears in between my messages. 
image

I could totally live with this though.

andersy005:
>  The progressbar stays in one piece, but appears in between my messages.

I will investigate this before merging, and will let you know if there's any workaround.

andersy005:
@jbusecke, I may have found a solution for your problem. 

When you get a chance, can you try replacing your print statements with

```python
print(..., end=' ') 
```

andersy005:
I am going to merge this for now. We can revisit it if any other issues show up

Issue #246, opened 2020-06-06
Problem when manually modifying dataframe


jbusecke:
I am trying to work with the CMIP6 data in the pangeo cloud working on ocean.pangeo.io . I would like to manually modify the dataframe of my esm_collection and then load only the dataset in the modified frame into a dict of xarray datasets.

I can reproduce the problem with a simple example:
```python
import intake
import pandas as pd
url = "https://raw.githubusercontent.com/NCAR/intake-esm-datastore/master/catalogs/pangeo-cmip6.json"
col = intake.open_esm_datastore(url)

cat = col.search(table_id=['Omon', 'Oyr'])

# now modify the dataframe manually
cat.df = cat.df.iloc[0:3, :]
```

When I call the `cat` object, the represention suggests that there are only 3 assets to load (as expeced) but that this would result in 700+ datasets:
![image](https://user-images.githubusercontent.com/14314623/83950649-623c4100-a7fa-11ea-82ad-75ea30d53592.png)

And indeed, when I call `to_dataset_dict`, it tries to load 754 datasets:
![image](https://user-images.githubusercontent.com/14314623/83950605-05d92180-a7fa-11ea-86d6-028b1fe9d79c.png)

I think I have successfully subsetted the dataframe in this way previously. Did something change or is there another recommended way to modify the dataframe?

```
Paste the command(s) you ran and the output.
If there was a crash, please include the traceback here.
```

### Output of ``intake_esm.__version__``

<details>

```
'2020.5.21.post2'
```
</details>

andersy005:
@jbusecke,

The safest way to create a new catalog object with a modified dataframe is to use the `from_df()` class method. This class method's signature looks like this:

```python
In [11]: cat.from_df?                                                           
Signature:
cat.from_df(
    df,
    esmcol_data=None,
    progressbar=True,
    sep='.',
    log_level='CRITICAL',
    **kwargs,
)
Docstring:
Create catalog from the given dataframe

Parameters
----------
df : pandas.DataFrame
    catalog content that would otherwise be in a CSV file.
esmcol_data : dict, optional
    ESM collection spec information, by default None
progressbar : bool, optional
    Will print a progress bar to standard error (stderr)
    when loading assets into :py:class:`~xarray.Dataset`,
    by default True
sep : str, optional
    Delimiter to use when constructing a key for a query, by default '.'
log_level : str, optional
    Level of logging to report, by default 'CRITICAL'

Returns
-------
[...]

PR #247, opened 2020-06-08
Add `df` property setter

andersy005:
Checklist

- [x] Tests added / passed

This PR adds a `df` property with its corresponding setter as suggested by @jhamman in https://github.com/NCAR/intake-esm/issues/246#issuecomment-640341221. The setter allows users to manually modify the dataframe if need be.

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -309,7 +309,7 @@ class esm_datastore(intake.catalog.Catalog):
     @df.setter
     def df(self, value):
         self._df = value
-        self._grouped, self._keys = _get_groups_and_keys(self._df, self.aggregation_info)
+        self._grouped, self._keys = _get_groups_and_keys(self.df, self.aggregation_info)
 
     def search(self, require_all_on=None, **query):
         """Search for entries in the catalog.
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -88,12 +88,12 @@ class esm_datastore(intake.catalog.Catalog):
 
         if isinstance(esmcol_obj, str):
             self.esmcol_data, self.esmcol_path = _fetch_and_parse_json(esmcol_obj)
-            self.df, self.catalog_file = _fetch_catalog(self.esmcol_data, esmcol_obj)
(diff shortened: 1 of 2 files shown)


andersy005:
The test failures appear to be unrelated to changes I made. I will investigate th e issue tomorrow, and will fix it in a separate PR.

intake_esm/core.py
```diff
@@ -707,3 +715,9 @@ def _make_entry(key, df, aggregation_info):
         name=key, description='', driver='esm_group', args=args, metadata={}
     )
     return entry
+
+
+def _get_groups_and_keys(df, aggregation_info):
```
jhamman, line 720: since this is only used inside the catalog class, maybe make this a class method. Something like:

```python
def _set_groups_and_keys(self):
    self._grouped = self._df.groupby(aggregation_info['groupby_attrs'])
    self._keys = list(grouped.groups.keys())
```

Then above, you would just call the method after setting `_df`:

```python
self._set_groups_and_keys()
```

andersy005: Done!

andersy005:
@jhamman, 

when you get a chance, please let me know whether you think this is ready :)

jhamman:
@andersy005 - can you take another look at the failing tests. They seem related to the changes here.

andersy005:
> can you take another look at the failing tests. They seem related to the changes here.

I have a hunch the failures aren't related to the changes made here because the same failures appeared in our [nightly builds](https://app.circleci.com/pipelines/github/NCAR/intake-esm/931/workflows/3517132c-a886-4971-9c09-3e98cfce1e13/jobs/3917/steps) for the last few days. I am looking into it though

andersy005:
It looks like `intake` recently changed the behavior of `Entry` objects in https://github.com/intake/intake/pull/490 and this broken some of our existing code/tests. We will need to adjust our code for Entry unification introduced in https://github.com/intake/intake/pull/490

andersy005:
I addressed the failing tests issue in #249. I am going to wait for #249 to be merged, and then I will rebase against master.

andersy005:
Thank you, @jhamman!

PR #249, opened 2020-06-09
Sync with intake's Entry unification

andersy005:
Checklist

- [ ] Tests added / passed

This PR adjust for changes in intake introduced in
intake/intake#490


andersy005:
For reason unbeknownst to me, the HTML repr appears to be broken as well. I can't tell whether it's related to the changes in intake or not. I will look into it tomorrow.

PR #259, opened 2020-07-07
Add support for wildcards in search() method

andersy005:
Checklist

- [x] Tests added / passed

This is meant to complement #236 

```python
In [4]: col = intake.open_esm_datastore(url)                                  

In [5]: col.search(long_name='^Wind.*direction$').df.head()                   
Out[5]: 
  component  ...                                               path
0       ocn  ...  s3://ncar-cesm-lens/ocn/monthly/cesmLE-20C-TAU...
1       ocn  ...  s3://ncar-cesm-lens/ocn/monthly/cesmLE-RCP85-T...
2       ocn  ...  s3://ncar-cesm-lens/ocn/monthly/cesmLE-CTRL-TA...
3       ocn  ...  s3://ncar-cesm-lens/ocn/monthly/cesmLE-20C-TAU...
4       ocn  ...  s3://ncar-cesm-lens/ocn/monthly/cesmLE-RCP85-T...

[5 rows x 9 columns]
```

Cc @markusritschel, @jeffdlb

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -756,7 +756,10 @@ def _is_pattern(value):
     if value_is_repattern:
         return True
     wildcard_chars = {'*', '?', '$', '^'}
-    return any(char in value for char in wildcard_chars)
+    try:
+        return any(char in value for char in wildcard_chars)
+    except TypeError:
+        return False
 
 
 def _flatten_list(data):
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -758,7 +758,7 @@ def _is_pattern(value):
     try:
         value_ = value
         for char in wildcard_chars:
-            value_ = value_.replace('\\{0}'.format(char), '')
+            value_ = value_.replace(f'\\{char}', '')
         return any(char in value_ for char in wildcard_chars)
     except (TypeError, AttributeError):
         return False
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -752,13 +752,15 @@ def _normalize_query(query):
 
 
 def _is_pattern(value):
-    value_is_repattern = isinstance(value, Pattern)
-    if value_is_repattern:
+    if isinstance(value, Pattern):
         return True
     wildcard_chars = {'*', '?', '$', '^'}
     try:
-        return any(char in value for char in wildcard_chars)
-    except TypeError:
+        value_ = value
+        for char in wildcard_chars:
+            value_ = value_.replace('\\{0}'.format(char), '')
+        return any(char in value_ for char in wildcard_chars)
+    except (TypeError, AttributeError):
         return False
 
 
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -758,7 +758,7 @@ def _is_pattern(value):
     try:
         value_ = value
         for char in wildcard_chars:
-            value_ = value_.replace(f'\\{char}', '')
+            value_ = value_.replace(fr'\{char}', '')
         return any(char in value_ for char in wildcard_chars)
     except (TypeError, AttributeError):
         return False
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -696,9 +696,9 @@ def _get_subset(df, require_all_on=None, **query):
             df[key].dtype, (np.object, pd.core.arrays.string_.StringDtype)
         )
         for val_i in val:
-            value_is_repattern = isinstance(val_i, Pattern)
-            if column_is_stringtype and value_is_repattern:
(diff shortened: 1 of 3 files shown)


intake_esm/core.py
```diff
@@ -756,7 +756,10 @@ def _is_pattern(value):
     if value_is_repattern:
         return True
     wildcard_chars = {'*', '?', '$', '^'}
-    return any(char in value for char in wildcard_chars)
+    try:
+        return any(char in value for char in wildcard_chars)
+    except TypeError:
+        return False
```
kmpaul, line 762: This does not test for escaped character sequences.  Presumably this is not a huge problem right now, but what if someone has a filename with a `$` in it?  Or a `*`?  Would it be safe to replace this test with:

```suggestion
    try:
        value_ = value
        for char in wildcard_chars:
            value_ = value_.replace(f'\{char}', '')
        return any(char in value_ for char in wildcard_chars)
    except TypeError:
        return False
```

tests/test_core.py
```diff
@@ -456,6 +456,7 @@ def test_normalize_query():
     [
         (2, False),
         ('foo', False),
+        ('foo\\*bar', False),
```
kmpaul, line 459: What about a single `\*`?  Doesn't that properly escape the `*`?

andersy005: Good catch! I overlooked this. This is now fixed.

tests/test_core.py
```diff
@@ -457,6 +457,7 @@ def test_normalize_query():
         (2, False),
         ('foo', False),
         ('foo\\*bar', False),
```
kmpaul, line 459: Is this correct?  `'foo\\*bar'` should reduce to `'foo*bar'` ...right?  Because `\\` is an escaped `\`?  Which would leave the `*`?

andersy005: As an update, this turned out to be harder than I anticipated. 

> Is this correct? 'foo\\*bar' should reduce to 'foo*bar' ...right?

It appears that the `\\` and `\` are being treated the same:

```python
In [40]: value = 'foo\\*bar'                                                                      

In [41]: char = "*"                                                                               

In [42]: value.replace(f'\{char}', '')                                                            
Out[42]: 'foobar'

In [43]: value = 'foo\*bar'                                                                       

In [44]: value.replace(f'\{char}', '')                                                            
Out[44]: 'foobar'

```

I am still looking into this, but I feel like I am starting to get confused ;)

kmpaul: Is this a Python thing or a `re` thing?  What about something like `r'\\*'`?

tests/test_core.py
kmpaul: ...See my comment below.  Apparently, in Python, `'\\*' == '\*'`.  So, I think you did it correctly.

tests/test_core.py
kmpaul: I was wrong.  You are correct.  Your original tests should be fine.

andersy005:
Thank you for the review, @kmpaul! I am going to merge this. Let me know if I missed anything and I'll address it in another PR.

wachsylon:
Hi @andersy005 ,
I cannot use wildcards in my ESM-Collections. Can you check it?
```
col=intake.open_esm_datastore("https://swift.dkrz.de/v1/dkrz_a44962e3ba914c309a7421573a6949a6/intake-esm/dkrz_cmip6_disk_netcdf_fromcloud.json")
col.search(variable_id="ta*")
```
Thanks in advance and best,
Fabi

andersy005:
Interesting... Which version of intake-esm are you using? You can get this info with the following

```python
In [1]: import intake_esm

In [2]: intake_esm.show_versions()

INSTALLED VERSIONS
------------------

cftime: 1.5.0
dask: 2021.07.1
fastprogress: 0.2.7
fsspec: 2021.07.0
gcsfs: None
intake: 0.6.2
intake_esm: 2021.1.15
netCDF4: 1.5.7
pandas: 1.3.1
requests: 2.26.0
s3fs: None
xarray: 0.19.0
zarr: 2.8.3

```

wachsylon:
INSTALLED VERSIONS
------------------

cftime: 1.5.0
dask: 2021.07.2
fastprogress: 0.2.7
fsspec: 2021.07.0
gcsfs: None
intake: 0.6.2
intake_esm: 2021.1.15
netCDF4: 1.5.6
pandas: 1.3.1
requests: 2.26.0
s3fs: None
xarray: 0.19.0
zarr: 2.8.3

wachsylon:
Sorry, it is working now... Maybe a wrong initialized kernel on my side.
Thanks for the quick answer.

PR #291, opened 2020-10-02
Added a function to search the time_ranges of the catalog

wachsylon:
Hi,
I am a scientific programmer at DKRZ and I really appreciate intake and intake-esm for the purpose of finding and accessing data.

An issue related to the subsetting in time ranges in the `time_range` column of CMIP catalogs is that the length of the string values vary. This code of this PR allows to search for all data that covers a time range defined by two time strings. 
We think that this will be very beneficial when working with CMIP catalogs.

A colleague of mine (Ludwig Lierhammer) has written this code snippet and asked me to share it.
I think the best place to implement that would be in your software in search.py so I tried to implement it there.
I tested the function itself however I do not know how to best call it in the intake workflow. Therefore, I do not have a test.

I am a beginner in python software managing. I try to follow your steps but I do also have not much time for caring for all the pull request tests especially since this code snippet needs some adaption.

And, an additonal comment:
I think that these 10 steps for a simple contribution to the repo takes too much time for an external programmer who has not much experience. This does not help if it is your intention to motivate the community to contribute to the repo. Idk how other repos handle it and I do understand that it is a part of guaranteeing stable code. For me, it is frustrating.
Cant that be done by automized CI tests?

Best regards,
Fabi

### DIFF
--- a/intake_esm/search.py
+++ b/intake_esm/search.py
@@ -5,7 +5,78 @@ from warnings import warn
 
 import numpy as np
 import pandas as pd
+from datetime import datetime
 
+def _search_time(df, needed_timerange):
+#Some checks for the format:
+    if len(needed_timerange) != 2 :
+        message="Timerange must have start and end values."
+        warn(message)
+        return pd.DataFrame(columns=columns)
+    if type(needed_timerange) == tuple:
+        needed_timerange = list(needed_timerange)
+    try:
+        int(needed_timerange[0])
+        int(needed_timerange[1])
+    except:
+        message="Timerange values must be convertable into integers."
+        warn(message)
+        return pd.DataFrame(columns=columns)
+
+#Functions used in the actual code:
+    def combine_alternately(S1, S2):
+        i = 0
+        while i < len(S2):
+            yield S1[i]
+            yield S2[i]
+            i = i + 1
+        yield S1[i]
+
+    def limit_format(fmt, date):
+        last_entry = int((len(date) - 2) / 2)
+        return fmt[:last_entry]
+
+    def select_fmt(date):
+        fmt = ['%Y','%m','%d','%H','%M','%s']
+        nondigits = [x for x in date if not x.isdigit()]
+        fmt = combine_alternately(fmt, nondigits) if nondigits else limit_format(fmt, date)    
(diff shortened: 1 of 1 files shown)


andersy005:
@wachsylon, Thank you for your patience and putting this PR together! Since it's been a while since you submitted this PR, I was wondering if this is something you are still interested in? 

I am asking because I think this is very useful functionality and would like to work with you to make it generic enough (the current implementation is a bit specific and tends to cover a **limited number** of use cases).

wachsylon:
Hi @andersy005 ,
ok, we could try.
> this is very useful functionality
cool.

> would like to work with you to make it generic enough
What do you have in mind? 

Best,
Fabi

aulemahal:
If I may, @andersy005 @wachsylon, we had the same issue and need when writing [xscen](https://github.com/Ouranosinc/xscen/). Our solution does not yet feel generic enough to be implemented here, but if you want to have a look a first version is here : https://github.com/Ouranosinc/xscen/blob/b261a04ed73e398a60a0632bdb29be324dc3f5b6/xscen/catalog.py#L899-L998

The idea is similar to this PR. You have a "date_start" and a "date_end" column in the catalog. For a given "period", the code returns the row where the `date_start - date_end` interval overlaps with that period. Because of the limitations of pandas <2, we had to use `pd.Period` objects in our catalogs and the code suffers a bit from this workaround. 

In addition to a simple "overlap", our function tries to guess the percentage of the period that is covered by the rows of the dataframe, so we only return the subset if a significant percentage is obtained. This has the restriction that it make sense if the rows of the dataset are not temporally overlapping, like for a single variable divided temporally in multiple files. A "full overlap" would often be too strict because of so many caveats (different calendars, imprecise date bounds).

I recently tried to use `datetime64[ms]` columns with pandas >= 2, which allows to simplify the function a bit and use more `pd.Interval` magic. It is here: https://github.com/Ouranosinc/xscen/blob/05054bfbf450c6b332e239e7866f766f51a47ed0/xscen/catalog.py#L892-L974.

There are still some caveats and questions to answer I think.
- How do we tell `intake-esm` which columns are the time bounds ?
- How do we solve the "coverage" issue neatly ?

PR #308, opened 2020-12-21
Update MPI catalogs for MISTRAL

aaronspring:
we moved our catalogs to a common path next to the cmorized data.


andersy005 (approved):
Thank you, @aaronspring

PR #393, opened 2021-11-02
Added detection of "*" to enable opening of multi-file datasets

agstephens:
## Change Summary

When `intake_esm.source._open_dataset()` is called, it now looks for `"*"` in the path and it will choose to open the dataset with `xarray.open_mfdataset()` instead of `xarray.open_dataset()`. This allows intake records to specify a glob pattern rather than an exact path.

## Related issue number

https://github.com/intake/intake-esm/issues/385

Some tests have been added in: `tests/test_source.py`. All the tests still run in my sandbox.

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/intake_esm/cat.py
+++ b/intake_esm/cat.py
@@ -315,7 +315,7 @@ class ESMCatalogModel(pydantic.BaseModel):
         results = search(
             df=self.df, query=_query.query, columns_with_iterables=self.columns_with_iterables
         )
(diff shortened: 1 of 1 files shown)


andersy005:
Thank you for putting this together, @agstephens! Sorry, I corrupted your main branch when I tried to push some changes to it. I am going to open another PR with your changes and I will ping you there

PR #403, opened 2021-11-23
Add `to_dask()` method to `esm_datastore`

### LINKED ISSUE #401: addition of `esm_datastore.to_dask()`
I often encounter the situation where my `search` on intake-ESM returns a result which is representable by a single dataset. I know that this is the case, because I craft my search in a way that this will be the case. However, extracting out the single dataset from the return value of `to_dataset_dict()` is always a bit lengthy.

As many intake libraries provide the `to_dask()` method to convert a result into an xarray dataset (which probably is a misnomer, but that's what people are using), I propose to add `to_dask()` to `esm_datastore` which should return a single dataset in case `to_dataset_dict()` would return exactly one item and fail in other cases.

I'd consider adding `to_dataset()` with the same functionality as above, but as other libraries seem to have settled on `to_dask()` already, I'd rather stay consistent than pedantic.

**Draft implementation**
```python
def to_dask(self, *args, **kwargs):
    if len(self) != 1:
        raise ValueError("not exactly one result")
    return next(iter(self.to_dataset_dict(*args, **kwargs).values()))
```

d70-t:
## Change Summary

This adds `esm_datastore.to_dask()` as discussed in #401 

## Related issue number

closes #401

## Checklist

- [x] Tests pass on CI

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -593,7 +593,7 @@ class esm_datastore(Catalog):
         if len(res) != 1:  # extra check in case kwargs did modify something
             raise ValueError(f'Expected exactly one dataset. Received {len(self)} datasets. Please refine your search or use `.to_dataset_dict()`.')
        _, ds = res.popitem()
-       return ds 
+       return ds
 
     def _create_derived_variables(self, datasets, skip_on_error):
         if len(self.derivedcat) > 0:
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -572,9 +572,7 @@ class esm_datastore(Catalog):
         self.datasets = self._create_derived_variables(datasets, skip_on_error)
         return self.datasets
 
-    def to_dask(
-        self,
-        **kwargs) -> xr.Dataset:
+    def to_dask(self, **kwargs) -> xr.Dataset:
         """
         Convert result to dataset.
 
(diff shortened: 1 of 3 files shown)


d70-t:
The failling CI seems to be unrelated to the change, is that correct?

intake_esm/core.py
```diff
@@ -572,6 +573,27 @@ def to_dataset_dict(
...
+        Parameters
+        ----------
+        all parameters are forwarded to :py:func:`~intake_esm.esm_datastore.to_dataset_dict`.
+
+        Returns
+        -------
+        :py:class:`~xarray.Dataset`
+        """
+        if len(self) != 1:  # quick check to fail more quickly if there are many results
+            raise ValueError('not exactly one result')
```
andersy005, line 591:
```suggestion
            raise ValueError(f'Expected exactly one dataset. Received {len(self)} datasets. Please refine your search or use `.to_dataset_dict()`.')
```

intake_esm/core.py
```diff
@@ -572,6 +573,27 @@ def to_dataset_dict(
...
+        res = self.to_dataset_dict(**{**kwargs, 'progressbar': False})
+        if len(res) != 1:  # extra check in case kwargs did modify something
+            raise ValueError('not exactly one result')
+        return next(iter(res.values()))
```
andersy005, line 595:
```suggestion
       _, ds = res.popitem()
       return ds 
```

intake_esm/core.py
```diff
@@ -572,6 +573,27 @@ def to_dataset_dict(
...
+
+        Returns
+        -------
+        :py:class:`~xarray.Dataset`
+        """
+        if len(self) != 1:  # quick check to fail more quickly if there are many results
+            raise ValueError('not exactly one result')
+        res = self.to_dataset_dict(**{**kwargs, 'progressbar': False})
+        if len(res) != 1:  # extra check in case kwargs did modify something
+            raise ValueError('not exactly one result')
```
andersy005, line 594:
```suggestion
            raise ValueError(f'Expected exactly one dataset. Received {len(self)} datasets. Please refine your search or use `.to_dataset_dict()`.')
```

andersy005 (commented):
@d70-t, thank you for putting this together... I left a few suggestions..

andersy005:
> The failling CI seems to be unrelated to the change, is that correct?

This is right... Something weird is going on... I will look into it and will fix it in a separate PR.

d70-t:
@andersy005 thank you for the review. I agree with all of the suggested changes.

andersy005:
Thank you for your contribution, @d70-t 🎉

Issue #407, opened 2021-11-24
Update Documentation


mgrover1:
Based on conversations regarding the recent changes, and user feedback, it would be good to do a review of our current documentation, finding areas for improvement. Examples include:
- [x] Add a "how to build a catalog"
- [ ] What is intake-esm doing with my data?
- [ ] What is a dictionary of datasets/collection?
- [x] How do I add a derived variable?
- [ ] How to disable aggregations in `to_dataset_dict()`
- [ ] How to modify `groupby_attrs` at runtime
- [ ] How to save a subsetted catalog
- [ ] Document the `skip_on_error` flag in `to_dataset_dict()`
- [ ] How to load a single key via the `cat[key].to_dask()`
- [ ] Guide on how to debug when things go wrong...
- [x] https://github.com/intake/intake-esm/issues/404
- [x] https://github.com/intake/intake-esm/pull/412: [furo theme](https://github.com/pradyunsg/furo)

Based on this [documentation system](https://divio-documentation-system.readthedocs-hosted.com/), we should cover
* Tutorials
* How to Guides
* Explanation
* Reference

mgrover1:
Related ESDS blogs
- ["Using Intake-ESM’s New Derived Variable Functionality" ](https://ncar.github.io/esds/posts/2021/intake-esm-derived-variables/)
- ["CESM2-LE Reproduction of Kay et al. 2015"](https://ncar.github.io/esds/posts/2021/kay-et-al-cesm2-le/)
- ["Reimagining Diagnostics Through the Use of the Jupyter Ecosystem"](https://ncar.github.io/esds/posts/2021/jupyter-based-diagnostics-overview/)
- ["Benchmarking Performance of History vs. Timeseries Files"](https://ncar.github.io/esds/posts/2021/benchmarking-history-timeseries-intake/)
- ["Regridding High Resolution Observations to a High Resolution Model Grid"](https://ncar.github.io/esds/posts/2021/regrid-observations-pop-grid/)
- ["Intake-ESM Holoviews Diagnostics"](https://ncar.github.io/esds/posts/2021/intake-esm-holoviews-diagnostics/)
- ["Using Intake-ESM to Analyze Data from CESM2-LE"](https://ncar.github.io/esds/posts/2021/intake-cesm2-le-glade-example/)
- ["Creating Model Documentation Using Jupyterbook and Intake-esm"](https://ncar.github.io/esds/posts/2021/model_documentation_jupyterbook/)
[...]

PR #409, opened 2021-11-24
Add Docs Based on ESDS Blog Posts

mgrover1:
Toward #407 

Include notebooks related to ESDS blog posts for documentation, including
- [x] Derived Variable Functionality
- [x] Building an Intake-ESM catalog from history files
- [x] Creating visualizations of Intake-ESM catalogs
- [x] Debugging the Intake-ESM process for reading in CMIP6 data


docs/source/user-guide/ecgtools-history-files-example.ipynb
```diff
@@ -0,0 +1,1337 @@
...
+    "pip install -e ecgtools\n",
+    "```\n",
+    "\n",
+    "This will install the package into the python environment you currently activated (for more on dealing with conda environments, check out the [faq page](https://ncar.github.io/esds/faq/#conda-environments)\n",
+    "\n",
+    "You will also want to install intake-esm, which you can install using `conda-forge`\n",
+    "\n",
+    "```\n",
+    "conda install -c conda-forge intake-esm\n",
+    "```\n",
```
andersy005, line 46: We should update this intro and remove references to the ESDS blog posts.

docs/source/user-guide/ecgtools-history-files-example.ipynb
```diff
@@ -0,0 +1,1337 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "czech-stylus",
+   "metadata": {},
+   "source": [
+    "### Understanding the Directory Structure\n",
+    "\n",
+    "The first step to setting up the `Builder` object is determining where your files are stored. As mentioned previously, we have a sample dataset of CESM2 model output, which is stored in `/glade/work/mgrover/cesm_test_data/`\n",
+    "\n",
+    "Taking a look at that directory, we see that there is a single case `b.e20.B1850.f19_g17.test`"
```
andersy005, line 74: It might be helpful to use publicly available datasets (e.g. the sample data within the repository) for this example if possible. It would make it easy for users to run and would make it easy to keep this notebook up-to-date by executing it during doc builds.

mgrover1: Good point!!

mgrover1: I just pushed the recent version which builds a catalog from data in the test directory

andersy005 (changes requested):
Thank you for putting together, @mgrover1! I left a few comments... Once these are addressed, let's merge this! 

- graphviz visualizaton notebook is broken
	- [x] Add python-graphviz to the doc requirements
- derived variable notebook
     - [x] Fix the missing data loading part. `ds` is used/mentioned without being defined. I presume `ds` is meant to the result of `.to_dataset_dict()`. 
     - [ ] Do you mind adding a section with a simple example that shows how to use `DerivedVariableRegistry.load(....)` to load a registry from a Python module? Here's an example from the unit tests: https://github.com/intake/intake-esm/blob/e367251e30b430a76409a96134eb2c3290e01802/tests/test_derived.py#L21

mgrover1:
@andersy005 - I added those missing/new sections you mentioned! When you get a chance, can you take a look and see if we can this merged soon?

andersy005:
Thank you, Max! 

The code for loading a registry from a Python module won't work unless you give it a path to a valid path. I don't think this has to be executable for demo purposes. How about making this code block non-executable by wrapping it in a markdown code block instead 

```python

dvr = DerivedVariableRegistry.load('my_registry') # my_registry points to a Python module (`my_registry.py`) file which has to be on Python's SYSPATH
````


Screen Shot 2021-12-06 at 12 18 13 PM

andersy005 (approved):
Thank you, @mgrover1! I will ping you for reviews when the docs rearrangement is ready

PR #416, opened 2021-12-14
Support mixed data formats

### LINKED ISSUE #414: Mixed formats in catalog
I work at a climate service provider. We do most of our work on local servers and a lot of the data is local. We want to use `intake_esm` to handle these in-house datasets, at two levels : the organisation-level and the project-level. Having a somewhat clean catalog at the project level is possible, but it is a complicated task at the org-level : i want to include a maximum of what we have in a single catalog and it's messy.

This means both netCDFs and zarr, as we are trying to move from the first to the second, but it's evidently impossible to completely abandon netCDF.

Currently, "assets.format" is mandatory and must be a single string ("netcdf" or "zarr"). The catalog and DataSource logic implies all url/paths use the same xarray "engine".

The esm-collection-spec has "assets.format_column_name" and it says :" The column name which contains the data format, allowing for variable data types in one catalog. Mutually exclusive with format." 
I'd like this to be reflected in `intake_esm`.

Having 2 catalogs. Or implementing some messy workaround in a custom subclass of `esm_datastore`.

**Additional context**
I can suggest a PR. My idea is the following:
- Making `Assets.format` optional. Add `None` as a valid enum item of `DataFormat`.
- Add `format_column_name` to `ESMDataSource` and pass it in `esm_datastore.__getitem__`.
- In `ESMDataSource.__init__`:  If `data_format` is not None,  set a new "_data_format" column to that, if it is None, rename the `format_column_name` to "_data_format".
- Use that new column in `ESMDataSource._open_dataset`.

That's the most elegant way I could think of, but I am open to suggestion if you are ok with this addition.

aulemahal:
## Change Summary

- Made `Assets.format` optional
- Added `format_column_name` to `ESMDataSource`'s init
- Format information is added a column on `ESMDataSource`'s internal df. `xarray_open_kwargs` are generated independently for each assets (instead of only once).

## Related issue number

- Fixes #414 
## Checklist

- [ ] Tests pass on CI

To make this PR simple, I decided to use an hardcoded name for the new column (`_data_format_`). Is this dangerous? It could be a column name stored in another variable, or the series could be independent from the main dataframe? The latter seems dangerous too.

I added a sample collection that has 4 assets of CMIP6. 2 assets are from Google Storage (zarr) and the 2 others  are local files. This has the bonus to also test that `to_dataset_dict` also works with remote and local assets _within the same dataset_.

### DIFF
--- a/intake_esm/cat.py
+++ b/intake_esm/cat.py
@@ -42,7 +42,7 @@ class Attribute(pydantic.BaseModel):
 
 class Assets(pydantic.BaseModel):
     column_name: pydantic.StrictStr
-    format: DataFormat
+    format: typing.Optional[DataFormat]
     format_column_name: typing.Optional[pydantic.StrictStr]
 
     class Config:
@@ -54,6 +54,8 @@ class Assets(pydantic.BaseModel):
         data_format, format_column_name = values.get('format'), values.get('format_column_name')
         if data_format is not None and format_column_name is not None:
             raise ValueError('Cannot set both format and format_column_name')
+        elif data_format is None and format_column_name is None:
+            raise ValueError('Must set one of format or format_column_name')
         return values
 
 
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -202,6 +202,7 @@ class esm_datastore(Catalog):
                     variable_column_name=self.esmcat.aggregation_control.variable_column_name,
                     path_column_name=self.esmcat.assets.column_name,
                     data_format=self.esmcat.assets.format,
+                    format_column_name=self.esmcat.assets.format_column_name,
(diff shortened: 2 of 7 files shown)


andersy005:
> To make this PR simple, I decided to use an hardcoded name for the new column (_data_format_). Is this dangerous?

I think this is good enough (It's similar to what we used to do before the re-write in https://github.com/intake/intake-esm/pull/368). 

> I added a sample collection that has 4 assets of CMIP6. 2 assets are from Google Storage (zarr) and the 2 others are local files. 

👍🏽  I think you forgot to commit the file though.

aulemahal:
Woops! And on the day I am working from home with the other machine shut down... I'll push them tomorrow.

PR #417, opened 2021-12-14
Search returns same class as self - allowing subclassing

### LINKED ISSUE #415: Allow subclassing `esm_datastore`
In the workflow I am building, which takes the form a package, I want to extend the `esm_datastore` class with some quite specific functionalities. Mostly things that don't really make sense in `intake_esm` directly. However, when subclassing `esm_datastore`, a call to  `search` returns a `esm_datastore` object because the class is explicitly initialized.

Replacing: 
https://github.com/intake/intake-esm/blob/e24e39ced1128f1b7cbc1edf9d467b1b8e301940/intake_esm/core.py#L347
with:
```python3
        cat = self.__class__({'esmcat': self.esmcat.dict(), 'df': esmcat_results})
```

A complicated override of `search` on my subclass. Since the introduction of the derived variable registry, this has been complicated and it's not only a matter of reconstructing a new object with the new `df`.

aulemahal:
## Change Summary
- In `esm_datastore.search` the result is instantiated from the same class as `self`, allowing subclassing.

## Related issue number

- Fixes #415

## Checklist

- [ ] Tests pass on CI

Should I add a unit test for this? It's not something I ever tested in other projects. I mean `intake_esm` does not have subclasses of `esm_datastore` thus testing this feature is somewhat out of scope? I can quickly add a basic one if you prefer.

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -344,7 +344,7 @@ class esm_datastore(Catalog):
                 .reset_index(drop=True)
             )
 
-        cat = esm_datastore({'esmcat': self.esmcat.dict(), 'df': esmcat_results})
+        cat = self.__class__({'esmcat': self.esmcat.dict(), 'df': esmcat_results})
         if self.esmcat.has_multiple_variable_assets:
             requested_variables = query.get(
                 self.esmcat.aggregation_control.variable_column_name, []
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -414,7 +414,6 @@ def test_to_dataset_dict_with_registry():
 
 
 def test_subclassing_catalog():
-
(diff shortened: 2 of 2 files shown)


andersy005 (approved):
Thank you for this contribution, @aulemahal! The CI failures are unrelated to this PR. 

> Should I add a unit test for this? It's not something I ever tested in other projects.

Feel free to add one if it doesn't require too much work. Otherwise, I'd be happy to merge this as is.

andersy005:
@aulemahal, let me know if you are planning to add the unit test otherwise I'm going to merge this later today.

andersy005:
Thank you, @aulemahal!

PR #443, opened 2022-02-01
Tuorial module for grabbing data

### LINKED ISSUE #437: Expose Intake-ESM test data for download
Understanding and learning intake-esm could be made much easier if the user had a sample dataset and catalog to reference.  It doesn't have to be "real" in any sense.  It could just be an example dataset, but I think there are pared down datasets in the `tests` directory of this repository that might fit the bill.  But I think it would be helpful if this dataset and corresponding catalog could be downloaded *with* intake-esm in the same way that `xarray.tutorial` provides an easy mechanism for downloading example datasets.

In the `tests` directory of this repository, the following catalogs and datasets appear to be already available as toy examples:
- `sample-collections/cesm1-lens-netcdf.csv`/`cesm1-lens-netcdf.json` + `sample-data/cesm-le/*.nc`
- `sample-collections/cmip5-netcdf.csv`/`cmip5-netcdf.json` + `sample-data/cmip/cmip5/*`
- `sample-collections/cmip6-netcdf-test.csv`/`cmip6-netcdf.json` + `sample-data/cmip/CMIP6/*`
- `sample-collections/multi-variable-catalog.csv`/`multi-variable-catalog.json` + `sample-data/cesm-multi-variables/*.nc`

I think an excellent solution might be to create an `intake_esm.tutorial` module that contains something like a `load_collection(name)` function (or `load_catalog(name)`?) that loads the above sample data and corresponding catalogs from GitHub, in the way that `xarray.tutorial.load_dataset(name)` does.  I would imagine that this function might return the path to the newly downloaded catalog (or the path to the existing catalog, if already downloaded?).

The obvious solution is to just tell people to `git clone` the repo and possibly reorganize the files so they are more easily "discoverable".  Perhaps that should be a first (simple) attempt.

jukent:
## Change Summary

Creates `tutorial.py` module to download files from GitHub URL for catalogues to be used in tutorial examples and easily accessible for new learners.

Updates __init__.py to access the `load_catalogue` fx

## Related issue number

Closes #437

## Checklist

- [ ] Tests pass on CI


jukent:
Looking at Xarray for inspiration: https://github.com/pydata/xarray/blob/main/xarray/tutorial.py

jukent:
@andersy005 during yesterday's meeting you suggested that this might not be the best method. Could you repeat your points here? What are all of our options?

jukent:
I was hoping to look at the collections already in the `/tests/sample-collections` folder - and trust that their datasets in `/tests/sample_data` are accessible.

First I wanted to make a tiny collection with only one entry, but for every collection I've tried such as: 

"""
url = "https://github.com/intake/intake-esm/blob/main/tests/sample-collections/cesm1-lens-zarr.json"
col = intake.open_esm_datastore(url)
"""

I get a `JSONDecodeError`. Do the collections need to be updated?

-----

I then planned to subset the collection to a single file with `.search()` and save it with `.serialize()`.
This new collection would be added to tests, called, searched, turned to a DataSet and compared to the DataSet loaded from just opening the one entry in the collection.

----

So as I understand it, what we want to be able to do is open a collection with a call like
`col = intake_esm.tutorial.cmip6`
instead of the 2 line
```
url = https://github.com/intake/intake-esm/blob/main/tests/sample-collections/cmip6-netcdf.json
col = open_esm_datastore(url)
```
(which doesn't work right now for me because of the JSONDecodeError)

The more I look at this, the less sure I am about how useful it is.

andersy005:
@jukent, one of the issues you are running into may have to do with the fact that inside those `JSON` files you are using, we are pointing to local (CSV) paths. As a result, intake-esm doesn't know how to retrieve the CSV files (hence the HTTP errors). If you inspect the tests, you will notice that when loading the catalogs in `sample-collections` directory, we don't use URLs but local paths.

jukent:
@andersy005 I did notice that and was about to try to use the `os.path.join()` method that is used to access the collections.

So I might need to make new collections anyway that do have URL paths? Or somehow tell Intake-ESM what the base URL is and append that to each path when downloading the file? I like the second option best.

PR #446, opened 2022-02-10
Subset derived variable registry only with used derived variables

### LINKED ISSUE #445: Don't include dependents when filtering the derived variable registry
Here's a quick checklist in what to include:

- [x] Include a detailed description of the bug or suggestion
- [x] Output of `intake_esm.show_versions()`
- [x] Minimal, self-contained copy-pastable example that generates the issue if possible. Please be concise with code posted. See guidelines below on how to provide a good bug report:

  - [Minimal Complete Verifiable Examples](https://stackoverflow.com/help/mcve)
  - [Craft Minimal Bug Reports](http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports)

    Bug reports that follow these guidelines are easier to diagnose,
    and so are often handled much more quickly.

### Description
If the Derived Variable Registry (DVR) includes a derived variable (dv2) that also happens to be a dependent for another derived variable (dv2), this var (dv2) gets included when the catalog search needs the second (dv1). Lets show an example:

```python3
import intake
import intake_esm
registry = intake_esm.DerivedVariableRegistry()

@registry.register(variable='FOO', query={'variable': ['FLNS', 'FLUT']})
def func(ds):
    return ds + 1

@registry.register(variable='FLNS', query={'variable': ['FOO', 'FLUT']})
def func2(ds):
    return ds - 1

cat = intake.open_esm_datastore(zarr_col_aws_cesm, registry=registry)
new_cat = cat.search(variable='FOO')
```
See how we provide a computation for `FOO` that requires `FLNS` and `FLUT` and also a computation for `FLNS` given `FOO` and `FLUT`. This makes a lot of sense and the goal here could be to fill missing variable that are not the same across datasets.

However, the current `search` method will first check for variable directly in the catalog, then parse the DVR to find if depends exist. In our case here, `FOO` is not available but `FLNS` and `FLUT` are. The problem resides in the subset of the DVR at the end of  `search` method because `FLNS` and `FLUT` were appended to the  `variables` list and thus the dependent variable for `FLNS` gets included in the final registry, even though it is not needed.

My suggestion would be to refactor (again) the derived variable search as to only include the dependent variable that did return results.
I can push a PR soon.

### Version information: output of `intake_esm.show_versions()`

<details>

Paste the output of `intake_esm.show_versions()` here:

```python
INSTALLED VERSIONS
------------------

cftime: 1.5.1.1
dask: 2021.12.0
fastprogress: 0.2.7
fsspec: 2021.11.1
gcsfs: 2021.11.1
intake: 0.6.4
intake_esm: 2021.8.17.post50
netCDF4: 1.5.8
pandas: 1.3.5
requests: 2.26.0
s3fs: 2021.11.1
xarray: 0.20.2
zarr: 2.10.3
```

</details>

aulemahal:
## Change Summary

When searching for variables, the subset of the derived variable registry is constructed from the dependent variables that returned results when their query was searched.

This mainly changes two things:
- If a derived variable has a dependent that also exists as a derived variable in the DVR, that second dv is nonetheless not included in the subset. Fixing #445.
- If a requested variable corresponds to a derived variable but none of the dependents were found in the catalog, that derived variable is not included in the subset.  I don't think this is a breaking change. That behaviour was not causing problems on my side, but I do think it's a bit cleaner this way?

## Related issue number
Fixes #445.

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/intake_esm/core.py
+++ b/intake_esm/core.py
@@ -330,10 +330,11 @@ class esm_datastore(Catalog):
         # This requires a bit of a hack i.e. the user has to specify the variable in the query
         derivedcat_results = []
         variables = query.pop(self.esmcat.aggregation_control.variable_column_name, None)
+        dependents = []
+        derived_cat_subset = {}
         if variables:
             if isinstance(variables, str):
                 variables = [variables]
-            dependents = []
             for key, value in self.derivedcat.items():
                 if key in variables:
                     res = self.esmcat.search(
@@ -346,8 +347,7 @@ class esm_datastore(Catalog):
                                 self.esmcat.aggregation_control.variable_column_name
                             )
                         )
-            # Update variable list with dependent variables.
-            variables = list(set(variables).union(dependents))
+                        derived_cat_subset[key] = value
 
         if derivedcat_results:
             # Merge results from the main and the derived catalogs
@@ -359,7 +359,7 @@ class esm_datastore(Catalog):
 
         cat = self.__class__({'esmcat': self.esmcat.dict(), 'df': esmcat_results})
         if self.esmcat.has_multiple_variable_assets:
-            requested_variables = variables or []
+            requested_variables = list(set(variables or []).union(dependents))
         else:
             requested_variables = []
         cat._requested_variables = requested_variables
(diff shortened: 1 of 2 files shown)


andersy005 (approved):
@aulemahal, Thank you for a thorough report in #445 and for your work on this! I appreciate it! 

This looks great to me 👍🏽

PR #449, opened 2022-02-22
Support for the grid mapping attribute and variable

RondeauG:
The `grid_mapping` attribute is used in CF Conventions to link a variable to its grid properties in the case of non-regular grids. These grids are very common for regional climate models. However, the current behaviour of `_open_dataset()` drops the grid_mapping variable (e.g. `rotated_pole`), since it is not recognized as a `requested_variables`

This small modification searches for the `grid_mapping` attribute within the `requested_variables` and if found, adds the related variable to the output.

### DIFF
--- a/intake_esm/source.py
+++ b/intake_esm/source.py
@@ -73,6 +73,7 @@ def _open_dataset(
             requested_variables = [requested_variables]
         variable_intersection = set(requested_variables).intersection(set(varname))
         variables = [variable for variable in variable_intersection if variable in ds.data_vars]
+        variables.extend([ds[v].attrs["grid_mapping"] for v in variables if "grid_mapping" in ds[v].attrs])
         ds = ds[variables]
         ds.attrs[INTAKE_ESM_VARS_KEY] = variables
     else:
--- a/intake_esm/source.py
+++ b/intake_esm/source.py
@@ -73,7 +73,7 @@ def _open_dataset(
             requested_variables = [requested_variables]
         variable_intersection = set(requested_variables).intersection(set(varname))
         variables = [variable for variable in variable_intersection if variable in ds.data_vars]
-        variables.extend([ds[v].attrs["grid_mapping"] for v in variables if "grid_mapping" in ds[v].attrs])
+        variables.extend(set([ds[v].attrs["grid_mapping"] for v in variables if "grid_mapping" in ds[v].attrs]))
         ds = ds[variables]
         ds.attrs[INTAKE_ESM_VARS_KEY] = variables
     else:
--- a/intake_esm/source.py
+++ b/intake_esm/source.py
@@ -73,7 +73,8 @@ def _open_dataset(
             requested_variables = [requested_variables]
         variable_intersection = set(requested_variables).intersection(set(varname))
         variables = [variable for variable in variable_intersection if variable in ds.data_vars]
-        variables.extend(set([ds[v].attrs["grid_mapping"] for v in variables if "grid_mapping" in ds[v].attrs]))
+        scalar_variables = [v for v in ds.data_vars if len(ds[v].dims) == 0]
+        ds = ds.set_coords(scalar_variables)
         ds = ds[variables]
         ds.attrs[INTAKE_ESM_VARS_KEY] = variables
     else:


andersy005 (approved):
Thank you for your patience, and contribution, @RondeauG! This looks good to me 👍🏽

andersy005:
@RondeauG, I also noticed this is your first PR here. Welcome to intake-esm!

intake_esm/source.py
```diff
@@ -73,6 +73,9 @@ def _open_dataset(
             requested_variables = [requested_variables]
         variable_intersection = set(requested_variables).intersection(set(varname))
         variables = [variable for variable in variable_intersection if variable in ds.data_vars]
+        variables.extend(
+            {ds[v].attrs['grid_mapping'] for v in variables if 'grid_mapping' in ds[v].attrs}
+        )
         ds = ds[variables]
```
andersy005, line 79: One quick comment/question... Presuming these `grid_mapping` variables are time invariant / static across datasets, should we cast these to `.coords` instead so as to avoid having them involved in the dataset concatenation/merging steps?

RondeauG: This is indeed a fundamental issue with those variables, but I think that this is an `xarray` issue rather than an `intake-esm` one. This is because despite having no dimension (see below for an example), `xarray` recognizes them as `data_vars` during `xr.open_dataset()`.

As for concatenation, the solution is to have ` xarray_combine_by_coords_kwargs = {data_vars: "minimal"}` during the call to `to_dataset_dict()`.

Example of a standard grid_mapping (using `ncdump`):

        char rotated_pole ;
                rotated_pole:grid_mapping_name = "rotated_latitude_longitude" ;
                rotated_pole:grid_north_pole_latitude = 42.5f ;
                rotated_pole:grid_north_pole_longitude = 83.f ;
                rotated_pole:north_pole_grid_longitude = 0.f ;
        float tasmax(time, rlat, rlon) ;
                tasmax:standard_name = "air_temperature" ;
                tasmax:long_name = "Maximum Near-Surface Temperature" ;
                tasmax:units = "K" ;
                tasmax:coordinates = "lon lat" ;
                tasmax:_FillValue = 1.e+20f ;
                tasmax:missing_value = 1.e+20f ;
                tasmax:realm = "atmos" ;
                tasmax:level_desc = "Height" ;
                tasmax:cell_methods = "time: max" ;
                tasmax:cell_measures = "area: areacella" ;
                tasmax:grid_mapping = "rotated_pole" ;

RondeauG: That being said, it is true that this might be a symptom of a more general issue with how `to_dataset_dict()` decides which variables to drop. There might be other cases similar to `grid_mapping` where a variable is used an a "info dump". I'm not sure how this can be easily addressed, however...

RondeauG: Coming back to this discussion, what do you think about using this instead:
`variables.extend(set([v for v in ds.data_vars if len(ds[v].dims) == 0]))`

This would grab all "informational" variables, instead of specifically _grid_mapping_.

RondeauG:
@andersy005 Sorry to tag you with this, but I just wanted to know if you have an ETA on this PR? Is there something missing and/or anything that I could do to help you with it? Thanks!

intake_esm/source.py
```diff
@@ -73,6 +73,9 @@ def _open_dataset(
             requested_variables = [requested_variables]
         variable_intersection = set(requested_variables).intersection(set(varname))
         variables = [variable for variable in variable_intersection if variable in ds.data_vars]
+        variables.extend(
+            {ds[v].attrs['grid_mapping'] for v in variables if 'grid_mapping' in ds[v].attrs}
+        )
```
andersy005, line 78: @RondeauG, how about we use your suggestion in https://github.com/intake/intake-esm/pull/449/files#r833298541?

```suggestion
       scalar_variables = [v for v in ds.data_vars if len(ds[v].dims) == 0]
       ds = ds.set_coords(scalar_variables)
```

this should cover all the issues (including the concatenation, etc...)

andersy005: once this is taken care of, this PR should be good to be merged

andersy005:
Thank you for this contribution and your patience, @RondeauG!

PR #453, opened 2022-03-10
Ensure fsspec storage options are propagated to `xr.open_dataset`

jukent:
## Change Summary

Fix bug that replaces storage options kwarg with an empty dictionary. Done with @andersy005 

## Checklist

- [x] Tests pass on CI


jukent:
Should we add a test that makes sure the 'storage_options` kwarg is maintained?

andersy005:
> Should we add a test that makes sure the 'storage_options` kwarg is maintained?

👍🏽  Let's add a simple test in https://github.com/intake/intake-esm/blob/main/tests/test_source.py

jukent:
> 👍🏽 Let's add a simple test in https://github.com/intake/intake-esm/blob/main/tests/test_source.py

SImple test added!

andersy005 (approved):
Thank you, @jukent!

Issue #456, opened 2022-03-11
Update Documentation to use new tutorial.py module for grabbing catalog URLs


jukent:
We now have curated small catalogs in the `tutorial-catalogs` directory from #454 and #455 that can be called via `intake_esm.tutorial.get_url()`. The documentation should be updated to use this method for some demonstrations.

In the same effort, the documentation can be updated to use "catalog" consistently (without misnamed references to "collections").

PR #458, opened 2022-03-14
Add `get_available_cats()` method to `tutorial.py`

jukent:
## Change Summary

Adding a `get_keys()` method to the `tutorial.py` module. This will help with the documentation by listing all of the available dataset catalogs to then choose from before calling `tutorial.get_url()`. 

Is it necessary to add a long-name associated with each of these keys?

Also corrects a spelling error on "catalog" in `tutorial.py` and adds a check that the `tutorial_catalogs` input kwarg is a dictionary.

## Related issue number

The need for this arose while working on #456 

## Checklist

- [x] Tests pass on CI


jukent:
Both `get_url` and `get_keys` allow the user to pass in their own dictionary of keys mapping to urls in the kwargs, instead of just the `tutorial_catalog` dictionary. There aren't tests checking for a bad dictionary input though.

andersy005 (approved):
This looks solid to me 👍🏽

thank you @jukent!

andersy005:
> Both get_url and get_keys allow the user to pass in their own dictionary of keys mapping to urls in the kwargs, instead of just the tutorial_catalog dictionary. 

my latest change restricts this to a default dict of values so as to reduce the complexity of the codebase. we can add this feature later when there's a demand for it.

PR #460, opened 2022-03-14
Change import to have configurable prefixes via `set_options`

aulemahal:
## Change Summary
Restructures the import of the default attribute names and prefixes in `source.py` so that they are actually configurable by patching the variables in `utils.py`.

## Related issue number
No issue sorry.

The problem was the `source.py` _imported_ the variables directly, this **copying** the python reference to the string. As strings  are non-mutable, someone that would want to customize the attribute names has to re-set the variable in the `utils` module (I think this is called "monkey-patching" ?). But that doesn't update the version in `source.py`. Here, the solution is to import the whole `utils` module object in `source`, so that the strings are queried as members of the module everytime they are needed, not only a import time.
## Checklist

- [x] Tests pass on CI

### DIFF
--- a/docs/source/reference/api.md
+++ b/docs/source/reference/api.md
@@ -50,3 +50,10 @@ For more details and examples, refer to the relevant chapters in the main part o
     :noindex:
     :special-members: __init__
 ```
+
+## Options for dataset attributes
+
+```{eval-rst}
+.. autoclass:: intake_esm.utils.set_options
+    :noindex:
+```
--- a/intake_esm/__init__.py
+++ b/intake_esm/__init__.py
@@ -8,7 +8,7 @@ from pkg_resources import DistributionNotFound, get_distribution
 from . import tutorial
 from .core import esm_datastore
 from .derived import DerivedVariableRegistry, default_registry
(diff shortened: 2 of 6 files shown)


aulemahal:
Oh! The way they were declared as constants made me think that was the goal. In our case, we found `intake_esm_attrs` to be kinda long to type. And, as we are subclassing the catalog class, the user has not explicit knowledge that `intake_esm` is used, which makes the name a bit weird. We are overriding it with "cat" for now.

I agree that this is not super clean. But there are only 3 variables... If you plan to add more user-controllable stuff, then yes an external package could be useful. For now, I feel it is not needed? I'm a dev at `xclim` and when we encountered the problem, we decided to go with the `set_options` function/context solution (like xarray), because it needed no extra package.

andersy005:
If you have time, let's go with the `options` dict approach used by xarray for now ( I don't envision other user-configurable settings in the future).

aulemahal:
@andersy005 Last commit pushed a simplified version of xarray's mechanic for options! I added the class to the API page and mentioned it in the docstring of `to_dataset_dict`.

PR #512, opened 2022-08-24
ENH: Add .to_datatree() method and remove to collection

### LINKED ISSUE #511: Integration with xarray datatree

A while ago, we added support for an `xcollection.Collection` object which contains a mapping of datasets. This object behaves like a dictionary with additional functionality. However, with the more recent development of xarray-datatree --which will eventually be integrated into Xarray proper--, we should transition to using `DataTree` instead of `Collection`. 

- https://github.com/intake/intake-esm/issues/382

- Provide a `cat.to_datatree()` method which returns an xarray `DataTree` object 
- Remove the `.to_collection()` method 

We currently provide a `cat.to_collection()` method which returns an `xcollection.Collection` object. 

**Additional context**

Cc @mgrover1

mgrover1:
## Change Summary
Adds a `catalog.to_datatree()` method, allowing for xarray datatree integration

## Related issue number

- Closes #511 

## Checklist

- [x] Tests pass on CI


mgrover1:
I tried adding the printed tree object to the docstring, but pre-commit struggled to tokenize that section... removing it for now.

mgrover1:
Alrighty - waiting for the tests to all pass and docs to build so I can take a look to double check things are working properly... but this should be ready for review @andersy005

andersy005:
thank you very much for working on this, @mgrover1!

intake_esm/core.py
```diff
@@ -7,7 +7,7 @@
 import pandas as pd
 import pydantic
 import xarray as xr
-import xcollection as xc
+from datatree import DataTree
```
andersy005, line 10: could we make datatree a soft/optional dependency?

andersy005: this would require importing it inside the `.to_datatree()` method and throwing an error when this method is called without `xarray-datatree`...

intake_esm/core.py
```diff
@@ -680,6 +678,10 @@ def to_collection(
             pr         (member_id, time, lat, lon) float32 dask.array<chunksize=(1, 600, 160, 320), meta=np.ndarray>
         """
 
+        # Set the separator to a / for datatree
+        self.sep = '/'
```
andersy005, line 682: we should make this change temporary and looking into resetting this to its original value before returning... 

```python
self.sep, old_sep = '/', self.sep
...
...
...


self.sep = old_sep
return ....
````

Comment (mgrover1) [reply intake_esm/core.py:682]: Yeah - that might be a better approach. Currently within datatree, I could not find a clean way of extracting the path to the datasets ("A/B/C/.."), and calling catalog.keys() was the only way to grab these strings nicely, and pass that into datatree to subset one of the datasets. For example:

```python
tree = collection.to_datatree()
ds = tree[collection.keys()[0]].ds
```

intake_esm/core.py
mgrover1: Yup! I can go ahead and add that

mgrover1:
@andersy005 I went through and made those suggested changes!

andersy005 (approved):
Looks solid to me 👍🏽

PR #569, opened 2023-02-03
Optional aggregation control

### LINKED ISSUE #568: Aggregation control still non-optional
Almost the same issue as #549. PR #551 fixed the pydantic model issue, however the code itself expected an `aggregation_control` object to exist, so the bug remains.

First, I'm not sure if `aggregation_control` _should_ be optional. When it is None, the keys of the dataset are simply all the columns joined together, including the path. There is also no `variable_column_name` and thus no derived variable registry and `ESMDataSource` can't select the expected variable.

I have a version where (AFAIK) `intake_esm` now works with `aggregation_control=None`, but I think an alternative would be to make it required. What do people think?

aulemahal:
## Change Summary
- Fixes #568 by introducing a bunch of if-else to work around cases where `cat.esmcat.aggregation_control` is None.

## Related issue number
- #568 

## Checklist

- [ ] Tests pass on CI

I added a few tests, but maybe I should add more to test most of the normal operations with a catalog missing the aggregation control.


mgrover1:
@aulemahal let me know when this is ready for review - happy to take a look. I think this approach makes sense.

mgrover1:
@aulemahal - following up here, is this ready for review?

aulemahal:
Sorry, I haven't had time to put on this in the last days. I still need to merge the conflicts and to add tests. I may have time in the next few days!

mgrover1:
No worries - take your time! I just wanted to check in and see if I could help with anything

intake_esm/core.py
```diff
@@ -101,6 +101,10 @@ def __init__(
         self._validate_derivedcat()
 
     def _validate_derivedcat(self) -> None:
+        if self.esmcat.aggregation_control is None and len(self.derivedcat):
+            raise ValueError(
+                'Variable derivation requires an `aggregation_control` to be specified in the catalog.'
```
mgrover1, line 106:
```suggestion
                'Variable derivation requires `aggregation_control` to be specified in the catalog.'
```

aulemahal:
This is finally ready for review @mgrover1 !

intake_esm/core.py
```diff
@@ -163,6 +167,10 @@ def keys_info(self) -> pd.DataFrame:
 
         """
         results = self.esmcat._construct_group_keys(sep=self.sep)
+        if self.esmcat.aggregation_control and self.esmcat.aggregation_control.groupby_attrs:
+            pass
+        else:
+            pass
```
andersy005, line 173: is this still necessary?

```suggestion
        if self.esmcat.aggregation_control and self.esmcat.aggregation_control.groupby_attrs:
            pass
        else:
            pass
```

mgrover1: I agree with @andersy005 - is this still neccessary?

aulemahal:
Here we go. While investigating the weird conditionnal, I realized that method was broken. Fixing it made me discover another deeper bug... So here we are with another implementation of the no-agg groups construction, which I believe is somewhat simpler.

mgrover1 (approved):
I like this approach!!

PR #589, opened 2023-03-29
Add `columns_with_iterables` parameter to `esm_datastore`

### LINKED ISSUE #587: Improve functionality for assets with Attributes containing iterables
It’s pretty common to have intake-esm assets that have multiple variables. The [current approach](https://intake-esm.readthedocs.io/en/stable/how-to/use-catalogs-with-assets-containing-multiple-variables.html) for dealing with these assets is to save an iterable of variables in the `variable_column` and pass a converter function to `intake_esm.esm_datastore`.

It’s also common to provide standard intake yaml descriptions of intake-esm catalogs (e.g. see [here](https://github.com/pangeo-data/pangeo-datastore/blob/master/intake-catalogs/climate.yaml)) to be passed to `intake.open_catalog`. The `args` to be passed to the `intake_esm.esm_datastore` driver are specified within the yaml file. It’s not possible to specify Python functions, meaning that assets with multiple variables can’t work by default.

Is it possible to fix this, so that an intake yaml description of an intake-esm catalog with multi-variable assets can work by default? E.g., a simple fix would be to add an optional `columns_with_iterables` parameter to the `intake_esm.esm_datastore` API that specifies the columns to convert with `ast.literal_eval`. Then, for example,

```yaml
sources:
  data0:
    args:
      obj: "/path/to/esm_cat.json"
      columns_with_iterables:
        - "variable"
        - "another"
    description: ""
    driver: intake_esm.esm_datastore
    metadata: {}
```
should provide a functional esm catalog when passed to `intake.open_catalog`.

I'm aware that users can use [`.configure_new`](https://intake.readthedocs.io/en/latest/catalog.html#local-catalogs) to override/add arguments for a source, but in this context this requires users to know which columns need converting which is not necessarily the case.

dougiesquire:
This PR adds an optional `columns_with_iterables` parameter to the `intake_esm.esm_datastore` API that specifies the columns to convert with `ast.literal_eval`. This enables intake yaml descriptions of intake-esm catalogs with multi-variable assets that work by default. See #587 for context/motivation.

Note, I also explored using intake [dataset transforms](https://intake.readthedocs.io/en/latest/transforms.html), but I don't think this provides quite the functionality we need.

Fixes #587 

## Checklist

- [x] Tests pass on CI


mgrover1 (changes requested):
I like this approach, especially with the error catching here. Would you be willing to add an example as well? So the docs reflect these new changes?

dougiesquire:
Thanks @mgrover1. Tests and docs added. Note, I had to pin `netcdf4<1.6.0` to get the tests to pass and docs to build. This is due to a change in 1.6.1 that seems to be causing issues all over the place, see e.g. https://github.com/Unidata/netcdf4-python/issues/1192

docs/source/how-to/use-catalogs-with-assets-containing-multiple-variables.md
```diff
@@ -12,7 +12,7 @@ kernelspec:
 By default, `intake-esm` assumes that the data assets (files) contain a single variable (e.g. `temperature`, `precipitation`, etc..). If you have multiple variables in your data files, intake-esm requires the following:
 
 - the `variable_column` of the catalog must contain iterables (list, tuple, set) of values (e.g. `['temperature', 'precipitation']`).
-- the user must provide a `converters` dictionary with appropriate functions for parsing values in the `variable_column` and/or any other column with iterables into iterables when loading the catalog. This is done via the `read_csv_kwargs` argument of the `open_esm_datastore` function.
+- the user must provide converters with appropriate functions for parsing values in the `variable_column` (and/or any other column with iterables) into iterables when loading the catalog. There are two ways to do this with the `open_esm_datastore` function: either pass the converter functions directly through the `read_csv_kwargs` argument, or simply specify the columns in `columns_with_iterables` parameter. The latter is a shortcut for the former. Both are demonstrated below.
```
mgrover1, line 15:
```suggestion
- the user must provide converters with appropriate functions for parsing values in the `variable_column` (and/or any other column with iterables) into iterables when loading the catalog. There are two ways to do this with the `open_esm_datastore` function: either pass the converter functions directly through the `read_csv_kwargs` argument, or specify the columns in `columns_with_iterables` parameter. The latter is a shortcut for the former. Both are demonstrated below.
```

docs/source/how-to/use-catalogs-with-assets-containing-multiple-variables.md
```diff
@@ -51,6 +51,16 @@ To confirm that intake-esm has loaded the catalog correctly, we can inspect the
 cat.esmcat.has_multiple_variable_assets
 ```
 
+Alternatively, we can simply specify the variable column name in the `columns_with_iterables` parameter:
```
mgrover1, line 54:
```suggestion
Alternatively, we can specify the variable column name in the `columns_with_iterables` parameter:
```

docs/source/how-to/use-catalogs-with-assets-containing-multiple-variables.md
```diff
@@ -69,6 +79,15 @@ dsets = cat_subset.to_dataset_dict()
 dsets
 ```
 
+## Why does `intake.open_esm_datastore` need the `columns_with_iterables` parameter?
+
+Why does intake `intake.open_esm_datastore` need the `columns_with_iterables` argument when we can achieve the same functionality with just `read_csv_kwargs`? Intake facilitates writing YAML descriptions of catalogs that can be opened with `intake.open_catalog`. These YAML descriptions include the information required to open the catalog: things like the catalog driver (`intake_esm.core.esm_datastore` in our case) and the arguments to pass to the driver to open the catalog. They can be included as entries in other catalogs enabling features like [catalog nesting](https://intake.readthedocs.io/en/latest/catalog.html#catalog-nesting). However, intake does not support Python function arguments like those we provided to `read_csv_kwargs` above so if we want a functional intake YAML description of an intake-esm catalog with multi-variable assets we need to use the `columns_with_iterables` arg instead. You can return an intake YAML description of an `esm_datastore` as follows:
```
mgrover1, line 84:
```suggestion
Why does intake `intake.open_esm_datastore` need the `columns_with_iterables` argument when we can achieve the same functionality with just `read_csv_kwargs`? Intake facilitates writing YAML descriptions of catalogs that can be opened with `intake.open_catalog`. These YAML descriptions include the information required to open the catalog: things like the catalog driver (`intake_esm.core.esm_datastore` in our case) and the arguments to pass to the driver to open the catalog. They can be included as entries in other catalogs enabling features like [catalog nesting](https://intake.readthedocs.io/en/latest/catalog.html#catalog-nesting). However, intake does not support Python function arguments like those we provided to `read_csv_kwargs` above so if we want a functional intake YAML description of an intake-esm catalog with multi-variable assets we need to use the `columns_with_iterables` argument instead. You can return an intake YAML description of an `esm_datastore` as follows:
```

mgrover1 (changes requested):
Just a few minor changes - I appreciate the contribution @dougiesquire !! Great work!

mgrover1 (approved):

andersy005:
thank you for this addition, @dougiesquire & @mgrover1!

dcherian:
Should this info be stored in the JSON instead? It's a property of the catalog, so the information can be specified at write-time instead of at read-time.

andersy005:
> Should this info be stored in the JSON instead? It's a property of the catalog, so the information can be specified at write-time instead of at read-time.

i'm definitely in favor of supporting this when it's defined in the JSON and adding it to the spec: https://github.com/intake/intake-esm/blob/main/docs/source/reference/esm-catalog-spec.md

PR #633, opened 2023-10-09
Use `chunks="auto"` by default in `xarray_open_kwargs`

### LINKED ISSUE #632: Change default `xarray_open_kwargs` to use `chunks="auto"`
Currently, by default, Intake-ESM opens files using `chunks={}`. 
For netCDF files:
- with xarray < v2023.09.0, this gives one chunk per file, which is sometimes a bad choice;
- with xarray >= v2023.09.0, this gives the file chunking, which is often a bad choice.

Default to `chunks="auto"` to use dask auto chunking. This should give better chunking in most cases.

Happy to submit a PR.

Passing the kwarg every time I use Intake-ESM

dougiesquire:
## Change Summary

Very small change to default to `chunks="auto"` rather than `chunks={}` when opening files with xarray. This is motivated by [recent changes in xarray](https://docs.xarray.dev/en/stable/whats-new.html#v2023-09-0-sep-26-2023) to the behaviour of `chunks={}`.

## Related issue number

Closes #632 

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/intake_esm/source.py
+++ b/intake_esm/source.py
@@ -19,7 +19,7 @@ def _get_xarray_open_kwargs(data_format, xarray_open_kwargs=None, storage_option
     xarray_open_kwargs = (xarray_open_kwargs or {}).copy()
     _default_open_kwargs = {
         'engine': 'zarr' if data_format in {'zarr', 'reference'} else 'netcdf4',
-        'chunks': {},
+        'chunks': 'auto',
         'backend_kwargs': {},
     }
     xarray_open_kwargs = (


dougiesquire:
Ah, unfortunately auto rechunking does not work with object dtype. This is probably too common for it to make sense to use `chunks="auto"` by default. Some options:

- try to use `chunks="auto"` and revert to `chunks={}` if that fails;
- use `chunks=-1`, which gives the same behaviour as pre-v2023.09.0 `chunks={}` (prior to v2023.09.0, these did the same thing);
- ditch this and stick with `chunks={}`

Thoughts @mgrover1, @dcherian?

mgrover1:
> Ah, unfortunately auto rechunking does not work with object dtype. This is probably too common for it to make sense to use `chunks="auto"` by default. Some options:
> 
> * try to use `chunks="auto"` and revert to `chunks={}` if that fails;
> * use `chunks=-1`, which gives the same behaviour as pre-v2023.09.0 `chunks={}` (prior to v2023.09.0, these did the same thing);
> * ditch this and stick with `chunks={}`
> 
> Thoughts @mgrover1, @dcherian?

I think sticking with `chunks={}` by default makes the most sense here still... any thoughts here @dcherian or @andersy005 ?

andersy005:
> > Ah, unfortunately auto rechunking does not work with object dtype. This is probably too common for it to make sense to use `chunks="auto"` by default. Some options:
> > 
> > * try to use `chunks="auto"` and revert to `chunks={}` if that fails;
> > * use `chunks=-1`, which gives the same behaviour as pre-v2023.09.0 `chunks={}` (prior to v2023.09.0, these did the same thing);
> > * ditch this and stick with `chunks={}`
> > 
> > Thoughts @mgrover1, @dcherian?
> 
> I think sticking with `chunks={}` by default makes the most sense here still... any thoughts here @dcherian or @andersy005 ?


i vote in favor of sticking with `chunks={}` unless implementing the work around above  ( * try to use `chunks="auto"` and revert to `chunks={}` if that fails;) doesn't result in complicated code :(

dougiesquire:
Thanks @mgrover1 and @andersy005. I agree with you both so I'll close this PR

PR #670, opened 2024-06-12
Added netcdf4 data format to support https access

hrhampapura:
I have added netcdf4 data format to the class DataFormat to support http access to netcdf4 files.


mgrover1 (approved):
Running the tests - once it is approved, we should be good to merge! Have you tested this out locally @hrhampapura ?

rbeucher:
@hrhampapura , could you provide an example / use case please?

PR #678, opened 2024-09-18
Fix how storage_options is passed to get_mapper.

garciampred:
## Change Summary

We identified a bug in the ESMCatalogModel.save method where storage_options were not being handled properly. The code now follows the same procedure used in the load method. if storage_options is None is transformed to an empty dict and double star expansion is used to expand it when passed to fsspec.get_mapper. The old way this storage_options where carried out to Aiobotocore.session is s3fs and makig it crash with this error:

```
E           TypeError: AioSession.__init__() got an unexpected keyword argument 'storage_options'
```
This fixed the issue.


andersy005 (approved):
thank you, @garciampred for this addition!

Issue #688, opened 2024-11-13
Next version release timeline


juntyr:
Thank you for creating this package! I wanted to ask when the next version of intake-esm will be released, since that one supports intake>=2.0.0. Thanks :)

rbeucher:
Hi @juntyr ,

We are looking into it!
Should not be too far away.

R

tlvu:
Hi, happy new year !  Any update on this release to support intake>=2.0.0 ?

rbeucher:
@charles-turner-1 ?

charles-turner-1:
We'll need to get a new version out for our own selfish purposes anyway (coordinate search)... I'd like to get the pin to `xr!=2024.10` in the requirements which is more of a stopgap solution fixed before we do so.

Realistically, allowing for other stuff that needs doing, we might be able to target early february?

andersy005:
> We'll need to get a new version out for our own selfish purposes anyway (coordinate search)... I'd like to get the pin to `xr!=2024.10` in the requirements which is more of a stopgap solution fixed before we do so.
> 

@charles-turner-1, the xarray version compatibility should be fixed now. i addressed it in https://github.com/intake/intake-esm/pull/696

charles-turner-1:
@andersy005 thanks very much!

Are there any other issues that you and/or @mgrover1 would want to get fixed before doing a new release?

charles-turner-1:
Awesome!

I'd like to work on #623 but I've done a little scoping it out and I think it'll take me a while to get up to speed with the parts of the codebase I'll need to understand to add date filtering, so unless you or @mgrover1 have any objections I'll trigger a release now.

juntyr:
Thanks, closed by https://github.com/intake/intake-esm/releases/tag/v2025.2.3

PR #683, opened 2024-10-29
Take2

charles-turner-1:
## Change Summary

- Updated intake version number to >2.0 in
    - `requirements.txt`
    - `ci/environment.yml`
    - `ci/environment-docs.yml`
- Specify xarray !=2024.10 (appear to be some issues related to datatree).
- Added explicit dask dependency (implicit dask dependency provided by intake < 2.0 removed in version 2.0)
 
## Related issue number

- Closes #649

## Checklist

- [x] Unit tests for the changes exist ( Dependency change - all tests passing).
- [x] Tests pass on CI (Read the Docs issue again).
- [x] Documentation reflects the changes where applicable ( I've grepped through the docs and can't find any reference to the intake version).

PR #689, opened 2024-11-15
Pin xr !=2024.10 in requirements.txt to match ci/environment.yml

charles-turner-1:
## Change Summary

xarray 2024.10 had a lot of changes to datatree - this breaks a number of tests. PR #683 pinned the xarray requirements in the CI to avoid this issue. 

I just forgot to update the xarray dependency to`xarray>=2022.06, !=2024.10` in `requirements.txt` to match CI in that PR.

Pending fix - I'd like to relax this before doing a new release (see #688)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,6 @@ intake>=2.0.0
 netCDF4>=1.5.5
 pandas>=2.1.0
 requests>=2.24.0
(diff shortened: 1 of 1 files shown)


charles-turner-1:
Going to fix the issue instead of just pinning dependency.

PR #690, opened 2024-11-17
feat: Support being able to use in-memory ESMCatalogModel instances

lewisjared:
## Change Summary

Support initialising the catalog using an in-memory instance of the data model. This avoids the need to serialise/deserialise if a temporary catalog is needed.

## Checklist

- [ ] Tests pass on CI


andersy005 (approved):
thank you for this addition!

Issue #705, opened 2025-03-05
Polars backend


charles-turner-1:
**Is your feature request related to a problem? Please describe.**
Some of our catalogues can take a long time to load - this has a negative impact on user experience.

**Describe the solution you'd like**
I've previously had a lot of success speeding dataframe operations up by doing everything in Polars, and then converting back to a pandas dataframe where necessary. I've done a very basic test, and with next to no work, this gives a ~2.5x speed increase: see diff & screencap.
```diff
-                df = pd.read_csv(
-                    cat.catalog_file,
-                    storage_options=storage_options,
-                    **read_csv_kwargs,
-                )
-            else:
-                df = pd.DataFrame(cat.catalog_dict)
+                read_csv_kwargs.pop('converters',None) # Hack, different args for polars
+                df = pl.read_csv(
+                    cat.catalog_file,
+                    storage_options=storage_options,
+                    **read_csv_kwargs,
+                ).to_pandas()
+            else:
+                df = pl.DataFrame(cat.catalog_dict).to_pandas()
```

<img width="823" alt="Image" src="https://github.com/user-attachments/assets/a6d65241-ff23-43ae-84a5-f631f29f9bdb" />

Making this work would require some code changes (currently breaks a few tests), but I'm fairly confident that I could do this relatively quickly, and that it wouldn't take a great deal of effort to take the performance benefit from ~2.5x to ~10-100x.

I would advocate doing any actual dataframe operations in polars, and then transforming back to a pandas dataframe when the user calls it to avoid any user facing changes.

NB. Polars has no dependencies, so this has the additional benefit of not making solving the environment a gigantic pain.

Is this something people would be interested in?

marc-white:
How different are the interfaces to pandas and Polars? Would "ripping off the bandaid" and forcing the users to start going in with Polars instead be a worthwhile effort?

I'd also be interested in:

- What's the time penalty for converting between Polars and pandas, and how does that scale with data size?
- What happens when a user gets one thing out of the datastore (as a pandas frame), and then passes it back in to the datastore as a comparator/selector for something else? Will that then need to be converted back into Polars internally?

charles-turner-1:
> How different are the interfaces to pandas and Polars? Would "ripping off the bandaid" and forcing the users to start going in with Polars instead be a worthwhile effort?

[...]

PR #709, opened 2025-03-21
Load with polars

charles-turner-1:
## Change Summary

- Change the `ESMCatalogModel.load()` method to use polars where possible, rather than pandas.
- Defer creation of pandas dataframe as long as possible.
- Change some implementation stuff surrounding accessing `columns_with_iterables`, etc, to account for deferred creation of polars dataframes.
- I've had to change a couple of tests, which relate to how pandas & polars deal with list columns differently. I think @dougiesquire wrote these - it would be good to get his opinion on whether the changes look like they've removed anything important. The approaches in the updated tests would now fail in the fall back to pandas method, so I'll add updated tests for those.
- Update requirements, etc.

Note: Polars struggles with `.bz2` compressed files. I haven't investigated this in depth, but my skim of the issue is that because Rust effectively treats strings of vectors of utf-8 characters & the bytecode compression makes this fall over. In these instances, we just fall back to the pandas read.

- ## Related issue number

I floated this idea in #705 . This PR just updates the `ESMCatalogModel.load()` to use polars wherever possible, rather than pandas. It won't exactly close it, since there are heaps of other operations - searching, filtering for unique values, etc - done using the pandas dataframe that could be moved over to using polars & should also see substantial performance boosts.

I'd suggest we make #705 a meta issue, and then gradually convert different functionality to polars to keep the PR's tractable. 

## Checklist

- [x] Tests pass on CI (N.B Segfault issues a la #697 are making this flaky)

## Performance

One of our largest datasets on Gadi, takes approximately 3-4s to load with the old backend. This is a `catalog.json`/`..csv.gz` combo:
```python
%%timeit -n 5 -r 5
intake.open_esm_datastore('/g/data/fs38/catalog/v2/esm/catalog.json')

3.54 s ± 97.8 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```
Using the updated polars based load doesn't change much if we keep using the zipped file:
```python
%%timeit -n 5 -r 5
intake.open_esm_datastore('/g/data/fs38/catalog/v2/esm/catalog.json')

3.15 s ± 218 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```
However, if we swap to an unzipped file, we get a ~100x speedup:
```python
%%timeit -n 5 -r 5
intake.open_esm_datastore("/scratch/tm70/ct1163/cmip6/catalog.json")

38.7 ms ± 15.7 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```
For an equivalent unzipped file, loading with pandas (current approach), we get:
```python
%%timeit -n 5 -r 5
intake.open_esm_datastore("/scratch/tm70/ct1163/cmip6/catalog.json")

2.79 s ± 134 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```

So something like a 75x performance improvement specifically in the case of unzipped files.


charles-turner-1:
This is not ready to merge - all tests are passing, but it seems to be the case that the readthedocs failure is due to my changes.

Will fix & add relevant tests.

charles-turner-1:
I've refactored a bit to make this less of a highly optimised, unreadable mess - moving dataframes out into a pydantic model, etc. Seems to have had negligible to positive effects on performance:
```python.
%%timeit -n 15 -r 15

# Using polars
intake.open_esm_datastore("/scratch/tm70/ct1163/cmip6/catalog.json")

18.7 ms ± 8.27 ms per loop (mean ± std. dev. of 15 runs, 15 loops each)
```

```python
%%timeit -n 5 -r 5

# Using pandas
intake.open_esm_datastore("/scratch/tm70/ct1163/cmip6/catalog.json")

3.08 s ± 165 ms per loop (mean ± std. dev. of 5 runs, 5 loops each)
```

Obviously `%%timeit` profiling has its flaws, but we're still sitting in the ~100x performance boost ballpark.

Think this is ready for review now.

mgrover1 (approved):
I appreciate the abstraction here - and defaulting to pandas where necessary! The speedups seem quite impressive - good to go on my end!

mgrover1:
@charles-turner-1 - feel free to merge when you are happy with it! I really appreciate the contribution here, and the massive speedup will be most welcome by the community!! 🚀

charles-turner-1:
@dougiesquire when you get a chance would you be able to look at the test I changed?

dougiesquire (approved):
Tests look sensible to me

PR #717, opened 2025-05-05
Add optional esmvalcore dependency & `to_iris method

### LINKED ISSUE #715: Facilitate creation of ESMValCore Dataset objects from datastores

See https://github.com/ESMValGroup/ESMValCore/pull/2690.

We're working on adding intake-esm support to ESMValCore/ESMValTool. I think one thing that would make this a lot more straightforward would be to add the ability to output an ESMValCore `Dataset` object from an `esm_datastore`. I'm envisaging something like

```python
esm_ds = intake.open_esm_datastore(...)
esm_ds = esm_ds.search(...)

dataset = esm_ds.to_dataset(...)
```

I think it would be wise to go one of two ways here: 
1. Make ESMValCore an optional dependency, in order to avoid making the dependencies of intake-esm too numerous. 
2. Double down on the intake plugin architecture and add an entrypoint that allows users to easily add extra methods to esm datastores - similar to the xarrays `register_dataset_accessor` system. Then we can stick the intake-esm / ESMValCore bridging code in a separate repo.

**Additional context**
https://github.com/ESMValGroup/ESMValCore/pull/2690, https://github.com/ACCESS-NRI/ESMValTool-workflow/issues/201#issuecomment-2846520663, https://github.com/ACCESS-NRI/ESMValTool-workflow/issues/202#issuecomment-2846514853

charles-turner-1:
## Changes
- Add optional imports following the way polars does it to keep overhead to a minimum.
- Add esmvalcore to list of optional imports.

## Related issue number

#715 

Closes #715 

## Checklist

- [ ] Tests pass on CI

### DIFF
--- a/intake_esm/__init__.py
+++ b/intake_esm/__init__.py
@@ -4,6 +4,7 @@
 
 # Import intake first to avoid circular imports during discovery.
 import intake
+import importlib
 
 
 from intake_esm import tutorial
@@ -12,3 +13,53 @@ from intake_esm.derived import DerivedVariableRegistry, default_registry
 from intake_esm.utils import set_options, show_versions
 
 from intake_esm._version import __version__
+
+_optional_imports: dict[str, bool | None] = {'esmvalcore': None}
+
+
+def _to_opt_import_flag(name: str) -> str:
+    """Dynamically create import flags for optional imports."""
+    return f'_{name.upper()}_AVAILABLE'
+
+
+def _from_opt_import_flag(name: str) -> str:
+    """Dynamically retrive the optional import name from its flag."""
+    if name.startswith('_') and name.endswith('_AVAILABLE'):
+        return name[1:-10].lower()
+    raise ValueError(
+        f"Invalid optional import flag '{name}'. Expected format: '_<import_name>_AVAILABLE'."
+    )
+
+
+__all__ = [
+    'esm_datastore',
+    'DerivedVariableRegistry',
+    'default_registry',
+    'set_options',
+    'show_versions',
+    'tutorial',
+] + [_to_opt_import_flag(name) for name in _optional_imports]
+
+
+def __getattr__(attr: str) -> object:
+    """
+    Lazy load optional imports.
+    """
+
+    if attr in (gl := globals()):
+        return gl[attr]
+
+    import_flags = [_to_opt_import_flag(name) for name in _optional_imports]
+
+    if attr in import_flags:
+        import_name = _from_opt_import_flag(attr)
+        if _optional_imports.get(import_name, None) is None:
+            _optional_imports[import_name] = bool(importlib.util.find_spec(import_name))
+            return _optional_imports[import_name]
+        else:
+            return _optional_imports[import_name]
+
+    raise AttributeError(
+        f"Module '{__name__}' has no attribute '{attr}'. "
+        f'Did you mean one of {", ".join(import_flags)}?'
+    )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,8 @@ import os
 
 import pytest
 
+import intake_esm
+
 here = os.path.abspath(os.path.dirname(__file__))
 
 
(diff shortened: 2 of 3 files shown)


charles-turner-1:
@rbeucher, this now works with the following script - loads an example dataset from the enso-recipes:

```python 

import intake
catalog = intake.cat.access_nri

# cat = cat.search(name='(?i)cmip.*', model = 'ACCESS-ESM1-5')
# yields cmip6_fs38

esm_ds = catalog['cmip6_fs38']
"""
This creates the same dataset as
model_datasets = {
"ACCESS-ESM1-5": 
    IntakeDataset(
    short_name='tos',
    project='CMIP6',
    mip="Omon",
    exp="historical",
    ensemble="r1i1p1f1",
    timerange="19790101/20190101",
    dataset="ACCESS-ESM1-5",
    grid="gn"
)}
when I was trying to add an IntakeDataset to esmvalcore.
"""

search = dict(
    variable_id='tos',
    table_id='Omon',
    experiment_id='historical', 
    member_id='r1i1p1f1',
    source_id='ACCESS-ESM1-5',
    grid_label='gn',
)

facets = dict(
    short_name='tos',
    project='CMIP6',
    mip="Omon",
    exp="historical",
    ensemble="r1i1p1f1",
    # timerange="19790101/20190101",
    dataset="ACCESS-ESM1-5",
    grid="gn"
)

esm_ds = esm_ds.search(
    variable_id='tos',
    table_id='Omon',
    experiment_id='historical', 
    member_id='r1i1p1f1',
    source_id='ACCESS-ESM1-5',
    grid_label='gn',
    version='v.*'
)

esmvalcore_dataset = esm_ds.to_iris(facets)
```

Is this the sort of functionality you had in mind/ we talked about?

Also cc'ing @bouweandela - this would go some way towards the esmvalcore/intake-esm integration by letting users dump an esmvalcore dataset out from an intake-esm search.

Todo:

- [ ] Auto-map between facets & search (this might be best handled in esmvalcore using the config & I think I can steal from Bouwe's old intake PR).
- [ ] Find out how to keep track of all searches made on a datastore so the user doesn't need to explicitly map facets.

I've added a `cmorizer` arg that I guess we'll need as we develop this further - as of right now I haven't got any idea how we might want to do that. Maybe @rhaegar325 might have some insight as to how this works for mopper? 

I have no idea if there is anything resembling a standard CMORizer API - if not, we might need to create something resembling one so that there's a plug & play way of cmorizing stuff in this functionality.

rbeucher:
> Thanks for cc'ing me! I'm not sure if integrating ESMValCore into intake-esm this is the right way around (I would personally integrate support for intake-esm into the [esmvalcore.dataset.Dataset.find_files](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset.find_files) method), but if it works for you then go for it.

I'm OK with that, but I agree it should be a `to_esmvalcore` function.

charles-turner-1:
> Thanks for cc'ing me! I'm not sure if integrating ESMValCore into intake-esm this is the right way around (I would personally integrate support for intake-esm into the [esmvalcore.dataset.Dataset.find_files](https://docs.esmvaltool.org/projects/ESMValCore/en/latest/api/esmvalcore.dataset.html#esmvalcore.dataset.Dataset.find_files) method), but if it works for you then go for it.

I'm planning on adding that too - the intake-esm branch/draft PR on the esmvalcore repo should have some code starting with that process already. Most of the functionality in this PR is implemented in the ESMValCore repo anyway.

This PR is mostly just low hanging fruit for our users, who are already going to be used to accessing data through intake catalogues.


> > if there is anything resembling a standard CMORizer API
> 
> Not yet, but there are plans to develop something like that as part of the upcoming Horizon Europe proposals. It will probably look something like `fix(ds: xarray.Dataset, identifier:str | None) -> xarray.Dataset` where `identifier` could be the `instance_id` (e.g. `CMIP6.CMIP.CSIRO.ACCESS-ESM1-5.historical.r1i1p1f1.Amon.pr.gn.v20191115`) for CMIP data and something else uniquely identifying the dataset for other data, or if not provided it could be read from the dataset attributes (assuming that those are correct).

Cool, that's really handy to know - thanks!

charles-turner-1:
Don't think this will be needed any more - gonna close.

PR #728, opened 2025-07-12
Add parquet driver

charles-turner-1:
## Change Summary

- Adds the ability to use parquet files
- Abstracts the reader/driver into a class
- [ ] Cleanup the nasty `if/else` chain - Moved to #729 
## Related issue number

N/A?

## Checklist

- [x] Tests pass on CI


mgrover1:
@charles-turner-1 - please let me know when this is ready for a review! Thanks for pushing this forward 🎉

andersy005:
great to see all the new additions and progress you’ve been making here :)

intake_esm/cat.py
```diff
@@ -227,7 +230,7 @@ def load(
         cls,
         json_file: str | pydantic.FilePath | pydantic.AnyUrl,
         storage_options: dict[str, typing.Any] | None = None,
-        read_csv_kwargs: dict[str, typing.Any] | None = None,
+        read_kwargs: dict[str, typing.Any] | None = None,
```
andersy005, line 233: given that is a major change that breaks backwards compatibility, can we gracefully introduce this in a way that doesn't break existing catalogs that have been using this kwarg?

i'm imagining something along these lines:

```python
read_csv_kwargs: dict[str, typing.Any] | None = None,
read_kwargs: dict[str, typing.Any] | None = None
```

and then in the function/method's body:

```python
if read_csv_kwargs is not None:
	warnings.warn("'read_csv_kwargs' is deprecated; use 'read_kwargs' instead", DeprecationWarning)

if read_kwargs is None:
	read_kwargs = read_csv_kwargs

else:
	raise ValueError("Can't use both 'read_csv_kwargs' and 'read_kwargs'")
```

charles-turner-1: Jeez, that was a real oversight on my part - thanks for catching this!

mgrover1: I agree with moving toward this approach!! This would break a number of things otherwise :)

charles-turner-1: Fixed it in the latest commits 🙏

charles-turner-1:
Okay, think this is more or less ready to go.

Stylistically I'm really not a fan of the if/else chain I've used to implement this, but I think it's going to be a bit of a pain to implement it 'more properly' - I'll open a separate PR doing so.

Issue #732, opened 2025-07-25
Iterable columns are better when hashable


aulemahal:
Hi!

WIth the recent implementation of a `polars` backend to read csvs, I realized that this repo's opinion is that iterable columns should be `list` objects. In our own layer added above `intake-esm` , we took for granted that variable columns would be tuples, thus being hashable and usable in more operations : `groupby`, `duplicated`, etc.

I would argue that hashable columns are better because of that. I see in commit a941786, for example, that the non-hashable nature of the column has been worked around by casting all cell into strings. While this work, I find it more complex and less readable than just using normal method like `drop_duplicates`. 

Sadly, the polars implementation removes all support for converters that could previously be passed to pandas. Also, it replaces the `ast.literal_eval` converter for it's own operation that assumes iterables are written as `"['foo', 'bar']"`.  In my catalogs, iterables are written as: `"('foo', 'bar',)"`'

I don't have an actual bug to submit, because `intake_esm` in it self doesn't break. However, in our tools building upon it, we rely quite heavily on patterns with `groupby` or `drop_duplicates` and this new behaviour breaks many functions.

Could we either:
- make `polars` an option to preserve previous functionality (custom converters, `literal_eval` as iterables conversion)
- allow both list and tuples constructs in the iterables conversion

or some other solution I haven't thought of ?

### Version information: output of `intake_esm.show_versions()`

<details>

```python
INSTALLED VERSIONS
------------------

cftime: 1.6.4
dask: 2025.7.0
fastprogress: 1.0.3
fsspec: 2025.7.0
gcsfs: None
intake: 2.0.8
intake_esm: 2025.7.9
netCDF4: 1.7.2
pandas: 2.3.1
requests: 2.32.4
s3fs: None
xarray: 2025.7.1
zarr: 2.18.7
```

</details>

charles-turner-1:
Hi Pascal, 

I think this should actually be a relatively straightfoward fix - I'm pretty sure this is just an oversight resulting from our uses cases (and the tests) not containing any examples of this behaviour breaking things. 

Can I bother you for a couple of snippets that are now broken in your workflow that I can use as the basis for new tests & I'll get the fixes in ASAP.

charles-turner-1:
With the ~fix in #734, I've just set the polars backend to always convert iterable columns to tuples. Not sure if this is completely safe - although it seems much more sensible than what's currently happening with iterable columns always being lists.

[...]

PR #734, opened 2025-08-05
Force iterable column to always be tuple

charles-turner-1:
## Change Summary

- Forces iterable columns to tuples when read with polars, following issue #732.  I can't really see any use case where we would prefer columns with iterables to be lists rather than tuples, I think it'll be simpler to just cast them to tuples everywhere in the in memory representation.
- Removes unnecessary try/except in `_search.py` which forcing iterable columns to tuples makes redundant.
- Add tests checking datatypes. @aulemahal if you could suggest some extensions for these tests so your use cases are all covered that would be great!

I think there might be a bit more work to do re. converters - but I think this should handle most of the problems that need fixing here. I also need to work out if there are any cases where forcing list columns to tuples is going to be problematic - struggling to think of any though - it's done eg. [in cat.py](https://github.com/intake/intake-esm/blob/d2e691afcd232a73f3181b07e7053c08d06f217f/intake_esm/cat.py#L344)

This will create a performance hit whenever we access the pandas dataframe associated with a catalog for the first time - realistically, the first `esm_datastore.search` or `esm_datastore.df` access is going to be when this penalty is paid. It should still be more performant than the pre-polars implementation, but will be a regression from current performance. I'd like to mitigate this performance hit going forwards if possible.

## Related issue number

See #732.

## Checklist

- [x] Tests pass on CI


mgrover1 (approved):
I like this approach!!

charles-turner-1:
Hey @aulemahal are you able to test xscen against the head of this branch and let me know if everything looks good before I merge? Cheers!

charles-turner-1:
No worries, sorry for causing you guys issues with the last release!

I see in the PR you tested on you mention categorical columns - I think it should be possible to pass categorical columns through to the polars reader using the `schema` kwarg in `read_kwargs`. In fact, I think this is probably something we'll want to implement on our own catalogs, so thanks for the prompt!

If you hit any performance bottlenecks, please reach out - improving performance for large catalogs is one of my major goals with this polars backend stuff, so if you encounter any regressions or memory usage issues please let me know!

PR #737, opened 2025-08-25
Change chunking default to auto for compatible xarray versions

charles-turner-1:
## Change Summary

- [x]  https://github.com/pydata/xarray/pull/10527

When the following is merged & released, we should be able to change the default chunking from `{}` to `"auto"`. 
This PR just preps intake-esm for that, slightly extending the proposed implementation from https://github.com/intake/intake-esm/pull/633/.
## Checklist

- [x] Tests pass on CI

### DIFF
--- a/tests/test_source.py
+++ b/tests/test_source.py
@@ -1,6 +1,7 @@
 import os
 import tarfile
 import tempfile
+from unittest import mock
 
 import dask
 import pytest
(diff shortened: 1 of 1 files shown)


charles-turner-1:
@mgrover1 Any chance you could take a look at this? think it's ready to go!

PR #742, opened 2025-09-07
Fix for polars 1.33.0 (dtype becoming mandatory)

charles-turner-1:
## Change Summary

- Specifies that iterable columns are decoded to the polars dtype `pl.List[pl.Utf8]` when opening catalogs with polars by default, allows user to specify dtype's via `schema_overrides`. (see https://docs.pola.rs/api/python/dev/reference/expressions/api/polars.Expr.str.json_decode.html)

## Related issue number

N/A

## Checklist

- [x] Tests pass on CI

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,7 +52,7 @@ jobs:
           python -m pytest
 
       - name: Upload code coverage to Codecov
-        uses: codecov/codecov-action@v5.5.1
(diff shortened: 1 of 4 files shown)


charles-turner-1:
@aulemahal I remember you saying you guys at ouranos are using categorical dtypes, are you able to take a skim through this PR & check that it doesn't look like it'll cause you any issues?

